PR #6, opened 2015-04-26
StdSchedulerMac: don't call setTolerance:

mistydemeo:
respondsToSelector isn't properly guarding against calling this method on unsupported platforms, possibly because it existed as a private method in OS X before 10.9. For example, attempting to build this on OS X 10.8 fails with the error:
```
no visible @interface for 'NSTimer' declares the selector 'setTolerance:'
```

Since the default tolerance is 0 anyway, this can be removed without adverse effect.


isilkor:
Merged. Thanks for the patch!

PR #12, opened 2016-02-09
Weather effects

lluchs:
See the [Trello card](https://trello.com/c/z9SXYu42/62-wettereffekte) for more information.

The only todo item still open is scaling down the splash graphics - I don't know which size to use there.


planet/Tests.ocf/Clouds.ocs/Map.c
```diff
@@ -0,0 +1,189 @@
...
+		map_size = [240, 200]; main_size = 80; nr_islands = RandomX(6, 7);
+	/* } */
+	/* if (SCENPAR_MapSize == 2) */
+	/* { */
+	/* 	map_size = [280, 220]; main_size = 90; nr_islands = RandomX(7, 8); */
+	/* } */
+	/* if (SCENPAR_MapSize == 3) */
+	/* { */
+	/* 	map_size = [320, 240]; main_size = 100; nr_islands = RandomX(8, 9); */
+	/* } */
```
isilkor: This code should either be made active, or removed entirely. (Also the comment style is strange.)

planet/Objects.ocd/Environment.ocd/Cloud.ocd/Script.c
```diff
@@ -249,28 +280,136 @@ private func Precipitation()
...
+	var x = x_orig, y = y_orig;
+	while (GBackSemiSolid(x, y - 1)) y--;
+
+	if (create_material)
+	{
+		InsertMaterial(Material(material_name), x, y);
+	}
+
+	// Some material combinations cast smoke
+	// TODO: Figure out some generic way to do this?
```
isilkor: Well, which material combinations are supposed to cast smoke? Right now, there's nothing that can be genericized because it looks like an arbitrary list.

planet/Objects.ocd/Environment.ocd/Cloud.ocd/Script.c
```diff
@@ -249,28 +280,136 @@ private func Precipitation()
...
-		mat = rain_mat_frozen;
+		return rain_mat_frozen;
 	else
-		mat = rain_mat;
-	// Create rain drop.	
-	CastPXS(mat, 1, 1, x, y);
+		return rain_mat;
+}
+
+local last_raindrop_color = -1;
```
isilkor: This should probably be initialized to `nil` instead of -1, because -1 (aka 0xFFFF'FFFF) is a valid color.

planet/Objects.ocd/Environment.ocd/Cloud.ocd/Script.c
```diff
@@ -249,28 +280,136 @@ private func Precipitation()
...
+		{
+			// Ice (-> hail) falls faster.
+			xdir *= 4;
+			ydir *= 4;
+		}
+
+		// Create rain particles.
+		var particle_name = "Raindrop";
+		var color = GetMaterialColor(mat);
+		if (color != last_raindrop_color)
```
isilkor: Why is your cache keyed on color instead of material?

lluchs: The particle definitions stored in the cache only depend on the color and would still work fine if we'd switch between materials of the same color (possibly Lava and DuroLava).

planet/Objects.ocd/Environment.ocd/Cloud.ocd/Script.c
lluchs: Yeah.  Those effects were created by Randrian. I guess I'll just move that check to a function to allow scenarios with custom materials to add to the list.

isilkor: Maybe you could use GetMaterialVal and check for Corrosive/Incindiary [sic]?

planet/Tests.ocf/Clouds.ocs/Map.c
lluchs: I should probably just draw a proper test map instead of copying Gem Grabbers...

planet/Objects.ocd/Environment.ocd/Cloud.ocd/Script.c
lluchs: Ah. I did that now.

PR #14, opened 2016-02-10
Use a proplist for optional Sound() arguments

lluchs:
Sound() has a lot of optional arguments which can be used independently. Most of the time, only a single one of these is needed, resulting in [oddly-looking code](http://docs.openclonk.org/en/sdk/script/SoundModifiers.html).

With this patch, a proplist can be passed instead, simulating named arguments. This pattern is common in other languages such as Ruby and JavaScript.

I also replaced some usages of Sound() with the new variant.

### DIFF
--- a/planet/Objects.ocd/Animals.ocd/PiranhaSmall.ocd/Script.c
+++ b/planet/Objects.ocd/Animals.ocd/PiranhaSmall.ocd/Script.c
@@ -79,7 +79,7 @@ private func UpdateVision()
 
 private func BiteEffect()
 {
-	Sound("Animals::Fish::Munch*", nil, nil, nil, nil, nil, 100);
+	Sound("Animals::Fish::Munch*", {pitch = 100});
 }
 
 local Name = "$Name$";
--- a/planet/Objects.ocd/Animals.ocd/Zaphive.ocd/Zap.ocd/Script.c
+++ b/planet/Objects.ocd/Animals.ocd/Zaphive.ocd/Zap.ocd/Script.c
@@ -170,7 +170,7 @@ private func CheckTurn()
 
 private func AngryBuzz()
 {
-	Sound("Animals::Zap::Zap?", nil,nil,nil,nil, nil, -Random(100));
+	Sound("Animals::Zap::Zap?", {pitch = -Random(100)});
 }
 
 /*-- Saving --*/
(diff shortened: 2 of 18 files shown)


SvenTwo:
I am not very happy with this change, beause it is inconsistent with most of our other functions. It also means constructing a lot of extra proplists on the fly. I would rather see named parameters implemented into the engine (i.e. Sound("Ding", player=1, global=true, volume=100)), like in Python. That is less verbose since you save an extra { }.

I think this would also be a relatively simple engine change to make since the parser could just reorder the arguments. Only question is whether evaluation of arguments should happen in the supplied order or in the order they are used in the function.

isilkor:
We don't currently document the order of parameter evaluation, so 
picking either is fine. However I don't think making the engine change 
is as easy as you claim, because of course functions implemented in C++ 
should work with named parameters too, and right now there's no way to 
get their parameter names - we could make that work with some 
preprocessor magic though.

lluchs:
If someone wants to implement real named parameters, I'd be happy with that as well. The long parameter list for Sound() has always been an issue though, and there's an easy fix available now.

I don't think creating lots of temporary proplists is an issue here, as Sound() isn't called that often (e.g. compared to CreateParticle() where you'd usually want to cache the proplist parameter).

SvenTwo:
Do we need preprocessor magic or could it also be template magic? E.g.:

bool Sound(StringPar<"name"> name, BoolPar<"global"> global, ...)

Of course the pre-processor could shorten it even more to avoid the duplication of the parameter name.

We can also use template specialization to define a GetTemplateName<T>-function to return NULL on plain parameter types, so you don't have to re-define all parameter names in one go, but just on some functions like Sound, Music, AddEffect, AddCommand, etc.

Another question is whether we also allow this for functions defined in script. If we do we would have to do name resolving at runtime because scripts can define the same function with different parameter names.

isilkor:
We'll have to allow it for script functions as well, otherwise people 
can't overload a function with named parameters anymore.

SvenTwo:
@isilkor They can. But it would have no effect on the parameter names.

When parsing parameters for a function call, the parser would simply search for an engine function of the same name and do its parameter sorting based on the parameter names. If you actually overload with different names, the order would be wrong of course.

isilkor:
I'll merge this because nobody is doing any work on named parameters at the moment, and in this case I value consistency less highly than the unreadable mess that functions with 10 optional parameters entail.

PR #16, opened 2016-02-11
SDL mouse event fixes

lluchs:
See http://bugs.openclonk.org/view.php?id=1682


isilkor:
Merged by Guenther in c332fe9.

PR #17, opened 2016-02-21
Improve gamepad support

lluchs:
[See the corresponding topic in the forums.](http://forum.openclonk.org/topic_show.pl?pid=31240)

I don't have Windows, so somebody should probably try to compile this there before merging.


lluchs:
>I don't have Windows, so somebody should probably try to compile this there before merging.
@ckanibal made building on Windows work now, so this is ready for merging now.

PR #18, opened 2016-02-24
Implement mouse grabbing for GTK on X11 (#1637)

lluchs:
The same thing should somehow also be possible using gdk_device_grab(), but I couldn't get it to work properly.

The pure GTK code would look something like this (without focus handling):

```cpp
void C4Window::GrabMouse(bool grab)
{
    GdkWindow *gdkwindow = gtk_widget_get_window(GTK_WIDGET(window));
    GdkDeviceManager *device_manager = gdk_display_get_device_manager(gdk_display_get_default());
    if (!device_manager) return;
    GdkDevice *pointer = gdk_device_manager_get_client_pointer(device_manager);
    if (grab)
    {
        gdk_device_grab(pointer, gdkwindow, GDK_OWNERSHIP_NONE, true, GDK_ALL_EVENTS_MASK, NULL, GDK_CURRENT_TIME);
    }
    else
    {
        gdk_device_ungrab(pointer, GDK_CURRENT_TIME);
    }
}
```

This wouldn't allow you to click outside the window, but the mouse cursor gets stuck when nearing the edge of the game window.

### DIFF
--- a/src/platform/C4WindowGTK.cpp
+++ b/src/platform/C4WindowGTK.cpp
@@ -378,6 +378,36 @@ static gboolean fullscreen_restore(gpointer data)
 	return FALSE;
 }
 
+static bool grab_mouse(GtkWidget *widget)
+{
+	// This should be possible with pure GTK code as well, using
+	// gdk_device_grab(). However, while gdk_device_grab() will prevent clicks
+	// outside of the game window, the mouse gets stuck near the edge of the
+	// window when trying to move outside.
+#ifdef GDK_WINDOWING_X11
+	Display * const dpy = gdk_x11_display_get_xdisplay(gdk_display_get_default());
+	Window xwindow = gdk_x11_window_get_xid(gtk_widget_get_window(widget));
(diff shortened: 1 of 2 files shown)


src/platform/C4WindowGTK.cpp
```diff
@@ -378,6 +378,34 @@ static gboolean fullscreen_restore(gpointer data)
 	return FALSE;
 }
 
+static bool mouse_was_grabbed = false;
```
aburgm: Can this be turned into a member variable of C4Window instead of a global variable? The `user_data` parameter in OnFocusInFS is a pointer to the C4Window instance.

aburgm:
It would be nice to have the explanation of why `gdk_device_grab` cannot be used also in the code as a comment.

src/platform/C4WindowGTK.cpp
lluchs: Sure. I used a global variable because the fullscreen_restore code above does the same thing.

lluchs:
I updated the commit.

PR #21, opened 2016-04-20
Use PCG as random number generator

lluchs:
PR is just there to track my todos:

- [x] Replace the particle RNG with PCG as well.
- ~~Use new `Random()` in `LoadScenarioSection()`~~ - probably not worth it on second thought.


isilkor:
lgtm after implementing the changes we've talked about in IRC.

PR #22, opened 2016-05-29
Add OpenClonk to StrategyGame and ArcadeGame categories.

pkern:
[Debian bug #825678](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=825678) suggested to modify OpenClonk's desktop file to add it to the StrategyGame and ArcadeGame categories to aid discoverability. Strategy sort of fits considering that it contains real-time strategy levels. Arcade seemed borderline but also felt true given that the tutorial and some other levels might be considered arcade.


isilkor:
Thanks for the PR; I've merged it and the changes should show up as soon 
as the mirror cron runs.

PR #23, opened 2016-05-29
Replace libupnp with miniupnpc

lluchs:
- The new code works with my router while libupnp didn't. :)

 - There are some unexplainable crashes in libupnp: #1640

 - Using miniupnpc seems to be less complex than libupnp.

 - Apparently, miniupnpc also works on Windows, so we may be able to use
   it for all platforms.

Disadvantage: UPnP queries aren't asynchronous anymore, but they seem to
be pretty fast (< 1 s).


src/network/C4Network2UPnPLinux.cpp
```diff
@@ -87,253 +36,100 @@ class C4Network2UPnPP: public C4InteractiveThread::Callback
...
 {
-	std::unique_ptr<Notify> notify(static_cast<Notify*>(pEventData));
+	if (!initialized) return; // Catches the case that UPnP initialization failed
 
-	// TODO: Should call a virtual method instead of dynamic_casting
-	NotifySearchResult* notify_search_result = dynamic_cast<NotifySearchResult*>(notify.get());
-	NotifyActionComplete* notify_action_complete = dynamic_cast<NotifyActionComplete*>(notify.get());
+	StdStrBuf eport, iport;
+	eport.Format("%d", static_cast<int>(mapping.external_port));
+	iport.Format("%d", static_cast<int>(mapping.internal_port));
```
isilkor: I still hate StdBuf, so I'd prefer `std::to_string(mapping.external_port)`

src/network/C4Network2UPnP.h
```diff
@@ -21,7 +21,7 @@
 
 class C4Network2UPnP
 {
-	struct C4Network2UPnPP *p;
+	class C4Network2UPnPP *p;
```
isilkor: If you change this you need to change src/network/C4Network2UPnPWin32.cpp too to avoid warnings (and potential link-time issues with MSVC because it uses mangling that breaks this)

src/network/C4Network2UPnPLinux.cpp
```diff
@@ -87,253 +36,100 @@ class C4Network2UPnPP: public C4InteractiveThread::Callback
...
-	std::string outgoing_address;
-	std::map<std::string, IGD> igds;
 	std::vector<PortMapping> added_mappings;
-	UpnpClient_Handle upnp_handle;
+
+	bool initialized = false;
+	char lanaddr[64];
+	struct UPNPDev *devlist = nullptr;
+	struct UPNPUrls upnp_urls;
+	struct IGDdatas igd_data;
```
isilkor: I assume these aren't forward declarations, so the `struct` isn't necessary.

src/network/C4Network2UPnP.h
lluchs: Oh, okay. I changed this to avoid a warning with the Linux build.

src/network/C4Network2UPnPLinux.cpp
lluchs: Right, I was thinking in C where the `struct` isn't optional.

lluchs:
Thanks for the comments, I updated the PR.

lluchs:
I added another commit which makes the UPnP actions asynchronous again. The `std::move`-in-lambda apparently needs C++14 but seems to be well supported in both gcc and clang.

PR #25, opened 2016-08-27
Added a Linux AppData file

Mailaender:
https://people.freedesktop.org/~hughsient/appdata/ has some details. See https://build.opensuse.org/request/show/423401 for an integration test.


isilkor:
Looks like none of our Linux people use a tool that parses these, but the format looks simple enough so I don't see the harm in adding it. Thanks for the patch.

PR #26, opened 2016-08-28
CMake improvements and fixes

susnux:
- Fix third-party libraries, they need to be explicit compiled as static libs.
  Some distributions use shared libraries as default preset in CMake.
- Fix some CMake checks (missing return in main function).
- Allow custom install directories (e.g. some distributions do not use share/games or usr/games and use share/ and usr/bin instead, so allow user to set a custom target, but still provide same default value as before).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,13 @@ CMAKE_DEPENDENT_OPTION(USE_WIN32_WINDOWS "Use Microsoft Desktop App User Interfa
 CMAKE_DEPENDENT_OPTION(USE_SDL_MAINLOOP  "Use SDL to create windows etc. Qt editor." ON "NOT USE_COCOA AND NOT USE_WIN32_WINDOWS" OFF)
 option(WITH_AUTOMATIC_UPDATE "Automatic updates are downloaded from the project website." OFF)
 
+
+if(UNIX)
+	# Allow setting custom directories on UNIX-like systems
+	set(OC_SYSTEM_DATA_DIR "${CMAKE_INSTALL_PREFIX}/share/games/openclonk" CACHE PATH "Path to openclonk datafiles on system.")
+	set(OC_SYSTEM_GAMES_DIR "${CMAKE_INSTALL_PREFIX}/games" CACHE PATH "Path to where to install games on the system.")
+endif(UNIX)
+
 set_property(GLOBAL PROPERTY USE_FOLDERS ${PROJECT_FOLDERS})
 
 ############################################################################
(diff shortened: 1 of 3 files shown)


Mailaender:
See it in action at https://build.opensuse.org/request/show/423579

isilkor:
- The installation code is currently only used on Linux (and maybe non-Darwin BSDs), so I think the new variable should not show up on Windows so people don't try to set it and have it do nothing.
- The STATIC change looks fine to me.
- ISO C++ provides that leaving `main()` without returning a value implicitly returns 0, and when I wrote the tests I've left the `return`s out since they aren't relevant to the test itself. But you're right that it's probably still a good idea to make them consistently return or not return.

susnux:
So do you would suggest to pack it into an `if (UNIX)` ?
Would make sense, from my point of view.

isilkor:
Either UNIX, or UNIX AND NOT APPLE. I don't know enough about Macs to 
say whether the setting makes sense there. @Meowtimer, @ckanibal?

ckanibal:
CMAKE_INSTALL_PREFIX will most likely not work for a end user on macOS due to system integrity protection. I'd rather have the application packaging script-fu working again (thing that creates .app-Files). 
Until then I suggest not using the _install_-method on macOS at all .

CMakeLists.txt
```diff
@@ -1459,7 +1466,7 @@ foreach(group ${OC_C4GROUPS})
 			DEPENDS "${native_c4group}"
 			VERBATIM
 		)
-		install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${group} DESTINATION share/games/openclonk)
+		install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${group} DESTINATION "${OC_SYSTEM_DATA_DIR}")
```
isilkor: Sorry for the late response, but this doesn't work on Windows (or rather, all non-UNIX platforms, of which there is only one we support) because OC_SYSTEM_DATA_DIR is unset, which makes CMake unhappy.

CMakeLists.txt
```diff
@@ -1474,7 +1481,7 @@ if (NOT APPLE)
 	install(CODE "execute_process(COMMAND update-desktop-database)")
 
 	# Install binaries
-	install(TARGETS openclonk DESTINATION games)
+	install(TARGETS openclonk DESTINATION "${OC_SYSTEM_GAMES_DIR}")
```
isilkor: Same as above.

jcaesar:
Since the commits are in master, I believe I can close this?

PR #27, opened 2016-09-17
Require Qt 5.4+ for editor — needs QOpenGLWidget

lheckemann:


isilkor:
Makes sense. Merged, thanks.

PR #28, opened 2016-10-01
Don't leak static proplists with cyclic references

aburgm:
If a local variable in a definition was set to a proplist inside the
Definition() callback, and that proplist contained cyclic references
then those references were leaked. Typically cyclic references for
script-created proplists are broken in
C4PropListScript::ClearScriptPropLists, however definition proplists
are changed to be static proplists in
C4PropList::FreezeAndMakeStaticRecursively.

To fix this, each script host maintains a list of proplists made static
by FreezeAndMakeStaticRecursively, and explicitly deletes all of these
proplists on Clear().

This leak also leads to an assertion failure inside
C4PropListScript::ClearScriptPropLists in debug mode, and can also be
observed by C4PropList::PropLists not being empty after game clear.

The definition in Objects.ocd/Helpers.ocd/UserAction.ocd constructs
cyclic proplists in its Definition() call. A simpler, more minimal way
to provoke the leak is the following (it provokes the leak but not the
assertion failure):

    local bla;

    func Definition(def)
    {
        bla = {};
        bla.test = { Name="Test222" , Options = { Name="Test333" } };
        bla.test.Options.Link = { Name="Test444", Blub=bla.test };
    }

@isilkor @scpmw @SvenTwo I'm not very familiar with this part of the codebase so would love someone looking over it to make sure it's reasonable. Is Günther on github?

### DIFF
--- a/src/script/C4Aul.cpp
+++ b/src/script/C4Aul.cpp
@@ -86,6 +86,18 @@ void C4AulScriptEngine::Clear()
 	GlobalNamed.SetNameList(&GlobalNamedNames);
 	delete pGlobalEffects; pGlobalEffects=NULL;
 	UserFiles.clear();
+	// Delete all global proplists made static (breaks
+	// cyclic references).
+	for (C4Value& value: OwnedPropLists)
+	{
+		C4PropList* plist = value.getPropList();
+		if (plist)
+		{
+			if (plist->Delete()) delete plist;
+			else plist->Clear();
+		}
+	}
+	OwnedPropLists.clear();
 }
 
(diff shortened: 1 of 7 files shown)


isilkor:
Do we need to box the proplists in a C4Value?

aburgm:
Yes, it is to make sure the engine does not delete the proplist elsewhere, which it does with non-cyclic-referenced proplists by just getting rid of its reference to them. We could possibly work around that by going through the new list beforehand, but it seems very fragile.

C4PropList refcounting does not work with a simple integer, but instead it uses a linked list of C4Values having references to that proplist. So it's hard to refcount it "manually" without boxing it in a C4Value.

PR #30, opened 2016-10-31
tools/convert_cr_to_oc.py : Added compatibility to Python 3

Fulgen301:


tools/convert_cr_to_oc.py
```diff
@@ -21,21 +22,21 @@ def convertname(root, files, name):
 		os.unlink(os.path.join(root, "Names.txt"))
 		r = "\"$Name$\""
 	else:
-		print root, name
-	for lang, localname in names.iteritems():
+		print(root, name)
+	for lang, localname in names.items():
 		if "StringTbl" + lang + ".txt" in files:
-			lines = open(os.path.join(root, "StringTbl" + lang + ".txt"),"rb").read().splitlines()
+			lines = open(os.path.join(root, "StringTbl" + lang + ".txt"),"r").read().splitlines()
```
isilkor: py3 still supports binary mode when opening files, and your changes will most likely break Windows compatibility...

tools/convert_cr_to_oc.py
```diff
@@ -21,21 +22,21 @@ def convertname(root, files, name):
...
 		else:
 			lines = ()
-		f = open(os.path.join(root, "StringTbl" + lang + ".txt"),"wb")
+		f = open(os.path.join(root, "StringTbl" + lang + ".txt"),"w")
 		for line in lines:
 			f.write(line)
 			f.write("\r\n")
```
isilkor: ...in particular in places like this, where the CRT translation layer will insert an additional `\r` in front of the `\n`, so you'd end up with `\r\r\n`.

tools/convert_cr_to_oc.py
Fulgen301: I tested the changes on Windows 10 x64. It worked. The point is, with the binary mode, you would need to use bytes-objects instead of str-objects which sucks with different encodings.

tools/convert_cr_to_oc.py
Fulgen301: I should not have checked it with Notepad >.<

isilkor:
I'm closing this PR for the time being since the added commits aren't related; feel free to reopen it from a different branch, though.

PR #31, opened 2017-01-05
IPv6 support

lluchs:
This is work in progress and not yet ready to merge.

Todos:

 - [x] Fix discovery/multicast
 - [x] Fix connections on `fe80::/64` link-local addresses (need to specify interface identifier).
 - [x] Fix existing IPv4 netpuncher
 - [x] IPv6 netpuncher for punching through the router's firewall?
 - [x] Use address from puncher to derive TCP address (for dual stack connections with port forwarding)
 - [ ] ~~Tests maybe? (Edit: No, existing tests are broken so tests are probably useless)~~
 - [x] Rebase and squash bug fix commits for bugs introduced in the first commit
 - [x] Replace all NULL with nullptr

Todo elsewhere:

 - [x] Make the league smarter about IPv6 (not sure yet how)


src/network/C4Network2Client.cpp
```diff
@@ -231,6 +231,8 @@ void C4Network2Client::AddLocalAddrs(int16_t iPortTCP, int16_t iPortUDP)
 					addr.SetPort(iPortUDP);
 					AddAddr(C4Network2Address(addr, P_UDP), false);
 				}
+				if (addr.GetScopeId())
```
lluchs: Can someone on Windows verify that this works correctly? E.g. by commenting out line 325 of C4Network2.cpp (above) and testing a local connection (should work even without IPv6 internet connectivity).

lluchs:
IPv6 multicast is now working in principle: Joining the multicast group works and I can see the discovery packages in Wireshark. However, a second engine on my system never receives anything from the other.

OC can apparently also use multicast for normal game traffic, but I don't think this is very useful - at least not until multicast traffic is actually routed over the internet. (There are a lot of posts about how IPv6 makes this possible in theory, but apparently noone every tried it out and complained that it doesn't work.)

isilkor:
Seems to be working (tested on a single machine only, might have to 
launch a couple VMs to test further).

lluchs:
Everything done!

src/netpuncher/C4PuncherPacket.h
```diff
@@ -29,6 +29,13 @@ enum C4NetpuncherPacketType {
 
 typedef uint32_t C4NetpuncherID_t;
 
+struct C4NetpuncherID {
```
isilkor: I'm not too happy with having two distinct types `C4NetpuncherID_t` and `C4NetpuncherID`

src/network/C4Network2.cpp
```diff
@@ -909,26 +928,46 @@ bool C4Network2::HandlePuncherPacket(C4NetpuncherPacket::uptr pkt)
...
 	}
 }
 
+C4NetpuncherID_t& C4Network2::getNetpuncherGameID(C4NetIO::HostAddress::AddressFamily family)
+{
+    switch (family)
+    {
+    case C4NetIO::HostAddress::IPv4: return NetpuncherGameID.v4;
+    case C4NetIO::HostAddress::IPv6: return NetpuncherGameID.v6;
+    default: assert(false);
```
isilkor: - Can you replace this with an assert that outputs a more useful message when it fails? E.g. `assert("Unexpected address family");` or something like that.
- This case still needs to return a value since in release builds we disable assertions and control will reach the end of the function.

src/network/C4NetIO.cpp
```diff
@@ -1769,20 +1814,21 @@ bool C4NetIOSimpleUDP::InitBroadcast(addr_t *pBroadcastAddr)
...
 	{
 		SetError("could not set mc ttl", true);
 		return false;
 	}
 
 	// set up multicast group information
 	this->MCAddr = *pBroadcastAddr;
-	MCGrpInfo.imr_multiaddr = static_cast<sockaddr_in*>(&MCAddr)->sin_addr;
-	MCGrpInfo.imr_interface.s_addr = INADDR_ANY;
+	MCGrpInfo.ipv6mr_multiaddr = static_cast<sockaddr_in6>(MCAddr).sin6_addr;
```
isilkor: I don't think we're guaranteed to have an IPv6 address in MCAddr at this point - [C4NetIOUDP](https://github.com/lluchs/openclonk/blob/18f474a6d16c1377759416c899f06ec3586cc799/src/network/C4NetIO.cpp#L2269) sets an IPv4 before calling this code.

src/network/C4Network2IO.cpp
```diff
@@ -460,28 +460,49 @@ bool C4Network2IO::InitPuncher(C4NetIO::addr_t nPuncherAddr)
...
-	PuncherAddr = nPuncherAddr;
+	switch (nPuncherAddr.GetFamily())
+	{
+	case C4NetIO::HostAddress::IPv4:
+	    PuncherAddrIPv4 = nPuncherAddr;
+	    break;
+	case C4NetIO::HostAddress::IPv6:
+	    PuncherAddrIPv6 = nPuncherAddr;
+	    break;
+	default:
```
isilkor: Maybe replace this with `case C4NetIO::HostAddress::Unknown: assert(!"insert meaningful error here");` so in case the IETF specifies IPv8 at some point at least we'll have a non-exhaustion warning that we can ignore

src/network/C4Network2IO.cpp
```diff
@@ -460,28 +460,49 @@ bool C4Network2IO::InitPuncher(C4NetIO::addr_t nPuncherAddr)
...
 
-void C4Network2IO::SendPuncherPacket(const C4NetpuncherPacket& p) {
-	if (!pNetIO_UDP || PuncherAddr.IsNull()) return;
-	pNetIO_UDP->Send(p.PackTo(PuncherAddr));
+void C4Network2IO::SendPuncherPacket(const C4NetpuncherPacket& p, C4NetIO::HostAddress::AddressFamily family)
+{
+	if (!pNetIO_UDP) return;
+	if (family == C4NetIO::HostAddress::IPv4 && !PuncherAddrIPv4.IsNull())
+		pNetIO_UDP->Send(p.PackTo(PuncherAddrIPv4));
+	else if (family == C4NetIO::HostAddress::IPv4 && !PuncherAddrIPv6.IsNull())
```
isilkor: I assume this should be `C4NetIO::HostAddress::IPv6`

src/network/C4Network2IO.cpp
```diff
@@ -1264,15 +1292,23 @@ void C4Network2IO::SendConnPackets()
 
 void C4Network2IO::OnPuncherConnect(C4NetIO::addr_t addr)
 {
-	// Sanity check
-	if (addr.GetFamily() != C4NetIO::HostAddress::IPv4)
-		return;
-	Application.InteractiveThread.ThreadLogS("Adding address from puncher: %s", addr.ToString().getData());
+	// NAT punching is only relevant for IPv4, so convert here to show a proper address.
```
isilkor: Does the IPv6 (firewall hole) puncher not connect back to the client then? Should there be an assertion (& early bail in release builds) to ensure this?

lluchs: It does connect back and the same code will be called. It just won't detect any new addresses, provided there's no IPv6 NAT. AddAddr checks for duplicates, so it's okay.

We could skip a bit of code, but I don't think it's worth the extra check here. This only happens once per game anyways.

src/network/C4NetIO.cpp
lluchs: Ah yes. I don't think that code is very useful in any case. Unicast works fine over the internet, so it is without doubt good enough for the local network. I left the code over there because maybe someone wants to have some fun with IPv6 multicast at some point. It seems to fail silently somewhere, but I can add a check here in any case.

lluchs:
I added some more commits which address @isilkor's comments. I'm probably going to squash most of them once everything is okay.

PR #32, opened 2017-01-30
HotIce: Fix weapon reload and hp-count on start

DasWipf:


lluchs:
This just removes the core part of the handicap feature. I suspect the real fix is to move the `InitPlayerRound` calls from `InitializePlayer` to `InitializePlayers. I'm going to try this.

PR #34, opened 2017-02-17
Pr 20170217

Fulgen301:
The commits include some fixes of bugs from https://trello.com/b/s3x6miCZ/openclonk-development (Board 8.0 Scenario Checklist)


isilkor:
Sven, Luchs, can you take a look at this? Afair you've been working with the guys on Trello.

SvenTwo:
Changes look good. I'm a bit worried why there's a var declaration on g_goal in Objects.c in the first place.

walachey:
b3b79e73dff19150258afc0d75f1d93ce322f58e looks odd. There is no comment about why it's done like that. (instead of just not creating the rule, that is).

Then there is one typo and still one forgotten English line in the German text. But I guess that can also be fixed post-merge.
(I just read through it and assume everything has been tested and works correctly ingame - especially the ```var g_goal``` thing)

PR #37, opened 2017-03-25
Rule_Relaunch

Fulgen301:
As discussed in #openclonk, I implemented a new relaunch rule which is able to spawn a clonk at the base, as Rule_BaseRespawn did, or at a specific location at the map. It also handles the relaunches.

(Note: If the changes get applied, DasWipf will have to modify his changes to the arena scenarios, as the new rule isn't compatible to them.)


isilkor:
This PR introduces a bunch of mixed whitespace, where some lines are indented with tabs and some are indented with 4 spaces. Please unify them to [use a hard tab for indentation](http://wiki.openclonk.org/w/C4Script_Style_Guidelines#Indents.2C_braces_and_spaces).

isilkor:
Tagging @MDT-Maikel and @SvenTwo for review because the changed files list them as author in the header.

lluchs:
Breaking all existing scenarios isn't great, even if there only are few. Would it be possible to have the old definition translate to the new rule? At least for the trivial case where the rule is just created without additional hackery.

Fulgen301:
@isilkor Ok, thank you for the notice.
@lluchs The only _really_ broken scenarios are the arena scenarios, as they have relaunch code in the scenario script. Unfortunately, I forgot to fix the BaseRespawn scenarios, which I will do now.

SvenTwo:
Looks good. I'm not too worried about minor changes to the scenarios. Relaunching is a big mess right now and has lots of code duplication in the goals and scenarios. We already have breaking changes with 7.0 scenarios because of the "Blunderbuss" rename. I assume the HotIce variants are not affected because they just use Melee and then do section reloading anyway, right? That doesn't leave too many scenarios to update.

Some remarks:
1. The rule introduces a few global functions; e.g. GetRelaunchCount. If you call these functions, they automatically create the relaunch rule. I think you wouldn't expect that just calling GetRelaunchCount enables relaunches. Maybe it could work like this:
```
var relaunch_rule = Rule_Relaunch->GetOrCreate();
var relaunch_count = relaunch_rule->GetRelaunchCount();
```
That would work without any global functions.


2. Base respawn has so much special handling. For example, inventory transfer is only implemented if you set base_respawn to true and free_crew is also ignored without base_respawn? I think base_respawn should only affect the relaunch positions and other attributes should be independent. So the position would be determined as:

a. GameCallEx to the scenario script other other goals (such as the parkour goal)
b. if a fails: A player base building if base_respawn is true
c. if a-b fail: A neutral respawn flagpole
d. if a-c fail: Random position


3. Parkour respawn is not implemented?


4. The relaunch rule could automatically create the clonk release rule (that allows you to restart from the rule menu). Alternatively, the two rules could just be merged into one. Scenario designers always forget to add that rule.

Fulgen301:
1) >I think you wouldn't expect that just calling GetRelaunchCount enables relaunches.
Well, the relaunches are stored within the rule, so without it calling GetRelaunchCount() does not make much sense ;)
Personally, I don't think using globals is a bad idea; if you want, I can modify them to library functions, as in the example.

3) Oh, yes. Skipped it because the parkour goals handles them and it is a special case - no one will use the parkour goal in a melee.
4) Ok, I'ill merge them.

PR #41, opened 2017-04-25
Automatic fixes using clang-tidy

tusharpm:
Following [the discussion on the forum](http://forum.openclonk.org/topic_show.pl?pid=33027), I ran a batch script to run clang-tidy with "modernize" checks enabled to make changes to the cpp files (no modifications to the header files, and files in thirdparty).
I have cleaned up the machine errors I could find, and as pointed out by others in the discussion.

Most of the changes can be summarized as follows:
- use C++ headers from the standard instead of the C-style ones.
- use bool literals (`false` or `true` instead of `0` or `1` respectively)
- use of null pointer (`nullptr` instead of `0`)
- change string-literals containing escape sequences to raw-string-literals.
- change for loops to range-for loops wherever applicable.
- use emplace (and similar) for containers wherever applicable.
- use `auto` instead of type names in select places (mostly range-for loop variable). I've reverted most usages of `auto` (501 to be exact) [as requested on the forum](http://forum.openclonk.org/topic_show.pl?pid=33014).

After the change, I've built the solution locally on Windows (MSVC), and Travis builds succeeded.


src/editor/C4ConsoleWin32.cpp
```diff
@@ -1139,12 +1139,12 @@ class C4ConsoleGUIPreviewWindow : public C4Window
 		pSurface = new C4Surface(&Application, this);
 	}
 
-	~C4ConsoleGUIPreviewWindow()
+	~C4ConsoleGUIPreviewWindow() override
```
tusharpm: Not to me. I guess, inherently, [I sided with clang-tidy](https://bugs.llvm.org//show_bug.cgi?id=30397), [you with the CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rh-override). As far as I can tell, a consensus wasn't reached before the discussion ended on either side. It is left for the programmers to choose their preference (`priority[0] = consistency;`).
Please let me know if you want me to revert these.

src/graphics/C4Shader.cpp
```diff
@@ -265,9 +265,9 @@ int C4Shader::ParsePosition(const char *szWhat, const char **ppPos)
 
 	// Lookup name
 	int iPosition = -1;
-	for (unsigned int i = 0; i < sizeof(C4SH_PosNames) / sizeof(*C4SH_PosNames); i++) {
-		if (SEqual(Name.getData(), C4SH_PosNames[i].Name)) {
-			iPosition = C4SH_PosNames[i].Position;
+	for (auto & C4SH_PosName : C4SH_PosNames) {
+		if (SEqual(Name.getData(), C4SH_PosName.Name)) {
+			iPosition = C4SH_PosName.Position;
```
tusharpm: Will do.

I'm still not used to the naming convention - especially using first letter capital for variable names, it'll take some time for me to adjust.

src/gui/C4KeyboardInput.cpp
```diff
@@ -602,10 +602,10 @@ C4CustomKey::C4CustomKey(const C4CustomKey &rCpy, bool fCopyCallbacks)
 	Name.Copy(rCpy.GetName());
 	if (fCopyCallbacks)
 	{
-		for (CBVec::const_iterator i = rCpy.vecCallbacks.begin(); i != rCpy.vecCallbacks.end(); ++i)
+		for (auto vecCallback : rCpy.vecCallbacks)
```
tusharpm: Will do.

src/gui/C4KeyboardInput.cpp
```diff
@@ -635,10 +635,10 @@ void C4CustomKey::Update(const C4CustomKey *pByKey)
 	if (pByKey->Codes.size()) Codes = pByKey->Codes;
 	if (pByKey->Scope != KEYSCOPE_None) Scope = pByKey->Scope;
 	if (pByKey->uiPriority != PRIO_None) uiPriority = pByKey->uiPriority;
-	for (CBVec::const_iterator i = pByKey->vecCallbacks.begin(); i != pByKey->vecCallbacks.end(); ++i)
+	for (auto vecCallback : pByKey->vecCallbacks)
```
tusharpm: Replaced all. 😄

src/gui/C4KeyboardInput.cpp
```diff
@@ -669,8 +669,8 @@ bool C4CustomKey::Execute(C4KeyEventType eEv, C4KeyCodeEx key)
 	// remember down-state
 	is_down = (eEv == KEYEV_Down);
 	// execute all callbacks
-	for (CBVec::iterator i = vecCallbacks.begin(); i != vecCallbacks.end(); ++i)
-		if ((*i)->OnKeyEvent(key, eEv))
+	for (auto & vecCallback : vecCallbacks)
```
tusharpm: Same.

src/graphics/C4Shader.cpp
lluchs: The worst thing about the variable naming is that it's not consistent at all, with Hungarian, capitals, camel case and snake case all over the place. The [style guide](http://wiki.openclonk.org/w/Style_Guidelines) mandates a specific style, but it often feels odd to follow that when all the surrounding code is in a different style.

isilkor:
Yeah, at one point I really want to change all of the ugly 90s hungarian 
style to something more readable, maybe one file at a time. While we're 
dreaming of a better world, I'd also like to unify function names and 
parameter names and members and all the other stuff that's accumulated 
over time.

src/editor/C4ConsoleWin32.cpp
isilkor: Ultimately I think the LLVM devs have a more compelling argument, so I think we should keep the `override`.

src/graphics/C4Shader.cpp
isilkor: [reposting this here where it belongs because my email reply went to the PR's general comments]

Yeah, at one point I really want to change all of the ugly 90s hungarian
style to something more readable, maybe one file at a time. While we're
dreaming of a better world, I'd also like to unify function names and
parameter names and members and all the other stuff that's accumulated
over time.

src/platform/StdSchedulerWin32.cpp
```diff
@@ -87,9 +87,8 @@ bool StdScheduler::DoScheduleProcs(int iTimeout)
 	// Iterate over the index because procedures may be added or removed during execution
 	// (If they are removed, we skip one execution, which doesn't really matter in practice)
 	auto tNow = C4TimeMilliseconds::Now();
-	for (size_t i_proc = 0u; i_proc < procs.size(); ++i_proc)
```
lluchs: This transformation is invalid, as the comment above explains. There are a couple more replacements like this one throughout the patch where I don't know whether iterator invalidation is an issue. I guess we'll see if anything crashes.

src/landscape/fow/C4FoWLight.cpp
```diff
@@ -305,9 +303,9 @@ void C4FoWLight::DrawFade(C4FoWDrawStrategy* pen, TriangleList &triangles) const
 {
 	pen->BeginFade();
 
-	for (TriangleList::iterator it = triangles.begin(); it != triangles.end(); ++it)
+	for (auto & tri : triangles)
 	{
-		C4FoWBeamTriangle &tri = *it; // just for convenience
```
lluchs: This comment is obsolete.

tusharpm: Removed the comment.

src/platform/StdSchedulerPoll.cpp
```diff
@@ -143,13 +143,12 @@ bool StdScheduler::DoScheduleProcs(int iTimeout)
 		bool any_executed = false;
 		auto tNow = C4TimeMilliseconds::Now();
 		// Which process?
-		for (size_t i = 0; i < procs.size(); i++)
+		for (auto proc : procs)
```
SvenTwo: May run into the same crash here (procs is potentially modified in the loop).

src/platform/StdSchedulerWin32.cpp
tusharpm: I assume that there is a possibility the collection is modified in the loop (`Execute` maybe?). If that is the case, all the useful iterators (including the past-the-end iterator) will be invalidated (because `procs` is a vector).
Making it a `list` may be advantageous **if the current element isn't removed**. This however loses random-access. A textual search for `procs` shows a handful of instances (I count 3) where `operator[]` is used, each of which can be systematically changed to something not requiring random-access. One thing that might get affected by the change is the locality of reference.
Please let me know what changes to make (or revert).

src/platform/StdSchedulerPoll.cpp
tusharpm: I observed the similarity. Please provide your thoughts on [my comment](https://github.com/openclonk/openclonk/pull/41#discussion_r113731696).

src/platform/StdSchedulerWin32.cpp
SvenTwo: I don't know if a schedule proc ever removes itself, but I find it very likely (if not, it may very well happen in the future).

I favor the indexed version (on a vector) because it's simple and it works no matter whether a schedule proc removes itself or another proc. The only downside is that removal offsets the index by one, so we miss execution of the following proc once. This should not matter in practice.

The specific loop did cause a crash in the past.

lluchs:
I still get some "inconsistent-missing-override" warnings from clang from the Qt files in src/editor. Did you miss these files?

tusharpm:
Evidently I did. I don't have Qt installed on my machine. It appears to be an optional dependency.
I'll install Qt and push another commit for those.

tusharpm:
I ran the script for the Qt files. There was no `override` in the diff.
Can you paste the whole warning text in a comment? It might give a hint about which file/line requires the change. It is possible that it is in one of the header files. If it is, it can be addressed in a different PR for header files, maybe? Please let me know if you want me to make those changes in this branch.

lluchs:
Ah, you didn't process header files at all? It's fine then. Not sure why I didn't notice this %)

lluchs:
Thanks, I squashed and merged everything.

PR #44, opened 2017-05-09
Automatic fixes using clang-tidy for header files

tusharpm:
A follow-up on a previous PR #41. The discussion in the forum can be viewed [here](http://forum.openclonk.org/topic_show.pl?pid=33086).
Run clang-tidy (without `auto`, `pass-by-value` and `using` checks) to fix the header files not modified in the previous PR.

Summary of the changes:
- C++11 member initialization.
- `nullptr` instead of `0` for pointers.
- `override` for functions declared `virtual` in base class.
- `default` trivial special member functions

This should cover the warnings mentioned by @lluchs [in this comment](https://github.com/openclonk/openclonk/pull/41#issuecomment-298649801).
After the change, I've built the solution locally on Windows (MSVC), and Travis builds succeeded.


SvenTwo:
Is it possible to let the tool generate assignments for the default initializers instead of braces? I always used assignment so far and it might make the code a bit inconsistent.

src/landscape/C4Particles.h
```diff
@@ -157,12 +157,12 @@ class C4ParticleValueProvider
 	bool IsConstant() const { return isConstant; }
 	bool IsRandom() const { return valueFunction == &C4ParticleValueProvider::Random; }
 	C4ParticleValueProvider() :
-		startValue(0.f), endValue(0.f), currentValue(0.f), rerollInterval(0), smoothing(0), keyFrameCount(0), valueFunction(0), isConstant(true), floatValueToChange(0), typeOfValueToChange(VAL_TYPE_FLOAT)
+		startValue(0.f), endValue(0.f), currentValue(0.f), rerollInterval(0), smoothing(0), keyFrameCount(0), valueFunction(nullptr), isConstant(true), floatValueToChange(nullptr), typeOfValueToChange(VAL_TYPE_FLOAT)
```
isilkor: These should presumably also be moved to the declaration.

isilkor: All of the particle classes seem to be missing the `modernize-use-default-member-init` check.

src/landscape/C4Texture.cpp
```diff
@@ -47,7 +47,7 @@ C4Texture::~C4Texture()
 }
 
 C4TexMapEntry::C4TexMapEntry()
-		: iMaterialIndex(MNone), pMaterial(nullptr)
+		: iMaterialIndex(MNone)
```
isilkor: `MNone` is a constant iirc, so could be moved to the declaration (unless of course its declaration isn't yet available in C4Texture.h)

src/lib/C4Random.h
```diff
@@ -20,7 +20,7 @@
 #ifndef INC_C4Random
 #define INC_C4Random
 
-#include <inttypes.h>
+#include <cinttypes>
```
isilkor: I think this is technically not equivalent (like in src/c4group/CStdFile.h); but should be fine in practice.

src/lib/StdBuf.h
```diff
@@ -46,7 +46,7 @@ class StdBuf
 	// Will take over buffer ownership. Copies data if specified.
 	// Note: Construct with Buf2.getRef() to construct a reference (This will work for a constant Buf2, too)
 	StdBuf(StdBuf & Buf2, bool fCopy = false)
-			: fRef(true), pData(nullptr), iSize(0)
+			: pData(nullptr)
```
isilkor: Are all these explicit `pData` initializations required? We're initializing in line 94.

src/network/C4NetIO.cpp
```diff
@@ -3923,9 +3901,7 @@ void C4NetIOUDP::DebugLogPkt(bool fOut, const C4NetIOPacket &Pkt)
 // *** C4NetIOMan
 
 C4NetIOMan::C4NetIOMan()
-		: StdSchedulerThread(),
-		iNetIOCnt(0), iNetIOCapacity(0),
-		ppNetIO(nullptr)
+		: StdSchedulerThread()
```
isilkor: The base ctor is called anyway and doesn't need to be listed explicitly; that way this entire function can be defaulted too.

tusharpm:
Yes. There's [an option](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-default-member-init.html#cmdoption-arg-useassignment) in `clang-tidy` to make it use the assignment syntax instead.
But there's a couple of things I'd like to point out before going that way:
 - [CppCoreGuidelines ES.23](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es23-prefer-the--initializer-syntax) suggests to prefer the `{}` syntax - for reasons listed in the linked section.
 - `clang-tidy` defaulted to braced-initialization, making it the more probable accepted form.

I agree that with the current change the code might look inconsistent; I propose replacing the current usages of the assignment operator (at least for data-members). This will make the code adhere to the core guidelines in the long run.
Please provide your thoughts regarding the same. The conclusion can possibly be added to the project's Style guidelines for future reference and guidance.

src/c4group/CStdFile.h
```diff
@@ -20,7 +20,7 @@
 #ifndef INC_CSTDFILE
 #define INC_CSTDFILE
 
-#include <stdio.h>
+#include <cstdio>
```
tusharpm: I agree, but this is something `modernize-deprecated-headers` did.
I guess a working build afterwards should be enough to ensure that nothing went wrong. Thoughts?

src/control/C4Control.h
```diff
@@ -54,18 +54,18 @@ class C4ControlPacket : public C4PacketBase
...
 
-	virtual void CompileFunc(StdCompiler *pComp);
+	void CompileFunc(StdCompiler *pComp) override;
 };
 
 #define DECLARE_C4CONTROL_VIRTUALS \
-  virtual void Execute() const; \
-  virtual void CompileFunc(StdCompiler *pComp);
+  void Execute() const override; \
+  void CompileFunc(StdCompiler *pComp) override;
```
tusharpm: Oops, it wasn't clang-tidy. That's my handiwork. Should I revert it from this PR?

src/c4group/CStdFile.h
isilkor: I don't think there's a single C++ compiler that implements <cstdio> by directly declaring the required members inside namespace std instead of just including <stdio.h> and `using` them into std, so this should be fine.

src/control/C4Control.h
isilkor: Keep it, it's fine.

src/control/C4Control.h
```diff
@@ -246,29 +241,28 @@ class C4ControlPlayerMouse : public C4ControlPacket // sync
...
 
-	C4ControlPlayerMouse() : action(CPM_NoAction), player(NO_OWNER), target_obj(0), drag_obj(0), old_obj(0) {}
+	C4ControlPlayerMouse() : player(NO_OWNER) {}
 	static C4ControlPlayerMouse *Hover(const C4Player *player, const C4Object *target, const C4Object *old_target, const C4Object *drag = nullptr);
 	static C4ControlPlayerMouse *DragDrop(const C4Player *player, const C4Object *target, const C4Object *drag);
 
 private:
-	int32_t action;
+	int32_t action{CPM_NoAction};
 	int32_t player;
```
tusharpm: I noticed that too. I don't know why it was left out (a few other places as well).
I think there's a few other tweaks to be made to my .clang-tidy config file. Maybe enabling `readability-redundant-member-init` will help.
I ran the same script again: it is still finding things to change. I'll upload the new commits after it finishes and I review the changes myself.

src/landscape/C4Particles.h
tusharpm: My copy of clang-tidy (on Windows) crashes consistently for some files in the project:
- src/landscape/C4Particles.cpp
- src/script/C4AulDebug.cpp

The Linux one succeeded though; I'm not sure why it didn't modify these. I'll run that again too.

src/lib/StdBuf.h
tusharpm: Second time's the charm :)

tusharpm: It didn't get modified in the second run also. Should I change it manually?

isilkor:
Merged in 233d7ed.

PR #45, opened 2017-05-11
Add expression for building with nix package manager

lheckemann:


tools/default.nix
```diff
@@ -0,0 +1,50 @@
...
+  hardeningDisable = "format";
+
+  nativeBuildInputs = with pkgs; [ cmake ];
+
+  buildInputs = with pkgs; [
+    SDL2 libvorbis libogg libjpeg libpng freetype glew tinyxml
+    openal freealut pkgconfig
+  ] ++ stdenv.lib.optional withEditor qt5.full;
+
+  cmakeFlags = [ "-DAudio_TK=OpenAL" ];
```
lluchs: I don't think this is necessary. cmake detects the OpenAL automatically and will prefer it even if SDL_mixer is available as well.

lheckemann: Oh yes, this just helped while I was trying to work out how to get it to find OpenAL (needed pkg-config!). I'll take it out.

lheckemann:
nixpkgs's cmake machinery doesn't allow changing the build type on the current release unfortunately. https://github.com/NixOS/nixpkgs/pull/25675 fixes this; I'll try adding `-DCMAKE_BUILD_TYPE=RelWithDebInfo` regardless and see how CMake responds to the multiple definitions, and if that doesn't work I'll just override the cmake hooks.

lheckemann:
I've made the change in https://github.com/lheckemann/openclonk/commit/19dd97142c0a7d0ce7f01d4a95b960783021d085. However, it fails to build: http://ix.io/ttN

isilkor:
We're doing link-time optimization in our release builds, which requires 
an ar/ranlib plugin (to properly handle the LTO/GIMPLE sections that gcc 
creates). gcc-ar/gcc-ranlib automatically use those plugins, but seem to 
not be available on your system, or at least not in a place where our 
(admittedly fragile) tests can find them. If gcc-ar and gcc-ranlib are 
available on nix, you can set AR_WRAPPER and RANLIB_WRAPPER to their 
paths, or alternatively you can disable LTO by passing 
-DUSE_GCC_STYLE_LTCG=Off to cmake.

Regarding the build type, I wonder if we should just add -g to the 
CFLAGS even for Release type builds, since it does use -O3 instead of 
-O2, and people can still strip the debugging symbols later if they 
don't want them. Or maybe we should use -O3 in RelWithDebInfo too.

lheckemann:
Builds with LTO now!

ben0x539:
fwiw, a previous release of openclonk is packaged in nixpkgs already: https://github.com/NixOS/nixpkgs/blob/ca6b9be5d197f19a59862f9c64b17fcda77e463d/pkgs/games/openclonk/default.nix

tools/default.nix
```diff
@@ -0,0 +1,65 @@
...
+
+  hardeningDisable = "format";
+
+  nativeBuildInputs = with pkgs; [ cmake ];
+
+  dontStrip = true;
+
+  buildInputs = with pkgs; [
+    SDL2 libvorbis libogg libjpeg libpng freetype glew tinyxml
+    openal freealut pkgconfig
```
ben0x539: Shouldn't pkgconfig go into nativeBuildInputs?

When I tried doing this, I also needed SDL_mixer, miniupnpc_2, mesa and xorg.xrandr, and didn't seem to need tinyxml, and the readme says we need zlib but I didn't try that.

tools/default.nix
```diff
@@ -0,0 +1,65 @@
...
+  ] ++ stdenv.lib.optional withEditor qt5.full;
+
+  cmakeFlags = [ "-DCMAKE_AR=${pkgs.gcc-unwrapped}/bin/gcc-ar" "-DCMAKE_RANLIB=${pkgs.gcc-unwrapped}/bin/gcc-ranlib" ];
+
+  preConfigure = ''
+    sed s/REVGOESHERE/''${gitRef:0:12}/ > cmake/GitGetChangesetID.cmake <<EOF
```
ben0x539: Can't we just add `"-DC4REVISION=${lib.escapeShellArg gitRev.rev}"` to `cmakeFlags`?

tools/default.nix
```diff
@@ -0,0 +1,65 @@
...
+
+  # Temporary measure until 28154f45a6 is in a release of nixpkgs
+  # Once this is the case, replace this with cmakeBuildType = "RelWithDebInfo"
+  configurePhase = ''
+    runHook preConfigure
+    fixCmakeFiles .
+    mkdir build
+    cd build
+    cmakeDir=..
+    cmakeFlagsArray+=("-DCMAKE_INSTALL_PREFIX=$prefix" "-DCMAKE_BUILD_TYPE=RelWithDebInfo" "-DCMAKE_SKIP_BUILD_RPATH=ON")
```
ben0x539: I think we might not need to override `configurePhase`, just appending `"-DCMAKE_BUILD_TYPE=RelWithDebInfo"` to `cmakeFlags` seems to work with me, overriding the `Release` value since `cmakeConfigurePhase` is always prepending.

tools/default.nix
lheckemann: The variable seems to get overridden unconditionally by the original GitGetChangesetID, so I don't think so.

tools/default.nix
lheckemann: It might work, but I'm not sure it's guaranteed behaviour on the part of cmake or nixpkgs. I'd prefer to go with the option that will definitely work.

tools/default.nix
lheckemann: Good point on pkgconfig.

Those that you suggest are missing are clearly not, otherwise it wouldn't build ;)

- SDL_mixer: we're using OpenAL as recommended by @lluchs on IRC. It replaces SDL_mixer and apparently works better.
- zlib doesn't need to be specified explicitly — I think libpng brings it in
- OpenGL is brought in by GLEW (so we don't need to specify mesa, which has the additional benefit of not making it specific to platforms that use mesa, i.e. this might build on OSX too)
- if tinyxml isn't provided by the system, it will use a vendored version (in thirdparty/tinyxml) which increases the size of the build
- xrandr is brought in by something else, probably SDL2
- miniupnpc is an optional dependency. It doesn't build correctly when it is available, I'll need to look into this in more detail.

lheckemann:
I'm aware of the previously packaged release. However, the build process and dependencies have changed so much since then that the expressions are almost completely unrelated to each other. In addition, this expression
 - Builds from git
 - Includes the hash in the version correctly, allowing compatibility with the snapshots

tools/default.nix
ben0x539: Ah, I think I didn't experience that because git wasn't available in my build so it didn't go into the GIT_FOUND branch.

Do you think we should set `C4REVISION_DIRTY` and `C4REVISION_TS` too?

tools/default.nix
ben0x539: Makes sense!

ben0x539:
Can we add an automated Nix build next? :)

tools/default.nix
lheckemann: Well git wouldn't be available in my build either since it's not in buildInputs. But the module will still set it to "unknown" as far as I can tell.

lheckemann: For C4REVISION_DIRTY, I don't see a convenient way to extract that information using nixpkgs's library and I don't think it's feasible without including the git repo in the source — at which point we could just use the normal GitGetChangesetID — but at the cost of many more unnecessary rebuilds. For C4REVISION_TS, it would be easier to get at it with just nix, but it's not implemented anywhere as far as I can see and requires decompressing the zlib-compressed commit file, which seems like more effort than it's worth.

lheckemann:
Using [nix-bundle](https://github.com/matthewbauer/nix-bundle), these could even be used on other distros! Although I'm not sure that solves an existing problem :p

lluchs:
Thanks!

PR #46, opened 2017-05-14
Cmake updates

tusharpm:
Modify a few things CMake can do better:
- get the time-stamp (in this case the year).
- fetch `googletest` (or generally allowing any external CMake project) using [DownloadProject](https://github.com/Crascit/DownloadProject) from @Crascit, which generalizes [the technique mentioned in `googletest`'s README.md](https://github.com/google/googletest/tree/master/googletest#incorporating-into-an-existing-cmake-project). The differences from master:
  - rename DownloadProject.CMakeLists.cmake.in -> DownloadProject.CMakeLists.txt.in
  - use the `@VAR@` syntax for `configure_file` with `@ONLY` option.
  - specify the build directory in the `cmake --build` command rather than using the `WORKING_DIRECTORY` option from `execute_process`.
- upgrade `googletest` to version 1.8.0.
- add tests to the `all` target and use `ctest` for running them.
  - building tests can be disabled with the `-DBUILD_TESTING=NO` option.

A local Windows (MSVC) build and Travis Linux builds succeeded and the tests passed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
       addons:
         apt:
           sources: ['ubuntu-toolchain-r-test']
-          packages: ['g++-4.9', 'build-essential', 'cmake', 'libfreetype6-dev', 'libgl1-mesa-dev', 'libglew-dev', 'libgtk-3-dev', 'libjpeg-dev', 'libpng-dev', 'libsdl2-dev', 'libupnp-dev', 'libxrandr-dev', 'x11proto-core-dev', 'zlib1g-dev', 'libalut0', 'libgtest-dev']
+          packages: ['g++-4.9', 'build-essential', 'cmake', 'libfreetype6-dev', 'libgl1-mesa-dev', 'libglew-dev', 'libgtk-3-dev', 'libjpeg-dev', 'libpng-dev', 'libsdl2-dev', 'libupnp-dev', 'libxrandr-dev', 'x11proto-core-dev', 'zlib1g-dev', 'libalut0']
       env: 
           - CCOMPILER=gcc-4.9 
           - CXXCOMPILER=g++-4.9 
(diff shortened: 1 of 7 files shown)


isilkor:
I'm okay with 4eba7f0, but I really don't like downloading any stuff from the configure script.

tusharpm:
Appreciate the review. Thanks.
As I like to see it, CMake is not exactly the same as a configure script, although it is usually used to do something very similar. For 13ff05e, there are two advantages of fetching googletest in CMake:
1. In googletest's readme: it is described as "a more robust and flexible approach". The best example I can give is it doesn't require the knowledge of which files to build/include from that repo.
2. This one is just for Windows users, but this way one can build and run the tests on their Windows machine in all configurations.

Also, the URL_HASH check optimizes the subsequent builds by skipping the file download if it matches (good for local builds - doesn't affect CI).

isilkor:
The problem I see with downloading additional files at compile (or 
configure) time is that you can't download the source tarball (and 
explicit dependencies) and then take the code with you on a train or a 
plane or wherever and start working on it, because the build will fail 
for lack of internet access. This is likely to also affect Debian 
builds, as I believe their packaging runs on servers that do not allow 
external network access. If we want to always have tests available, I'd 
prefer shipping a convenience copy of gtest, or using a git submodule.

tusharpm:
Would adding `find_package(GTest OPTIONAL)` and using this as fallback be better? That would however require googletest to be installed independently using the platform's native package manager.

isilkor:
I've merged 4eba7f0 (as 9020f3c) but I've decided I won't merge 13ff05e. I see the advantages, but (and this is of course just a personal opinion) I'm very unhappy with doing any network access from CMake; doing this from a helper script - which could download the dependencies and then call CMake - would be fine, but a chore to maintain, considering that if we want to go that route, it should of course not just download one arbitrarily selected library, since all major Linux distributions maintain packaged versions of GTest, and Windows doesn't come with several other libraries we depend upon either.

tusharpm:
Thanks for the merge.
For the other dependencies (especially for Windows), I had tried [a similar approach in my first PR](https://github.com/openclonk/openclonk/pull/39/files#diff-473f2985d91d820c1be94e0218e3d22f). The major difference is that required the dependencies to be built before the configure script. This doesn't, but imposes a different restriction on the dependencies to make their CMake files "embed"-friendly. I assumed we could start with the test framework and build the solution further from there. I understand your concern that just one library would look out of place.
For the submodule approach in #49: IMO, they are tougher and messier to maintain/upgrade. But that's a personal opinion. I've experience with a project where we tried it out and changed to this mechanism.

PR #47, opened 2017-06-03
Rain drop effect

Fulgen301:
The rain drop effect from https://trello.com/c/gbwfJhuh/143-grafischer-wassertropfen-effekt

### DIFF
--- a/planet/Objects.ocd/Environment.ocd/Stalactite.ocd/Script.c
+++ b/planet/Objects.ocd/Environment.ocd/Stalactite.ocd/Script.c
@@ -21,7 +21,7 @@ public func Construction()
 		sibling = TransformBone(bone, transformation, 1, Anim_Const(1000), sibling);
 	}
 	
-	AddTimer("DoWaterdrop", RandomX(36 * 10, 36 * 40));
+	AddRainDropEffect(nil, 100, "Water", 2);
 }
 
 private func Hit()
@@ -31,8 +31,8 @@ private func Hit()
 	{
 		Size = PV_KeyFrames(0, 0, 0, 100, PV_Random(3, 5), 1000, 3),
 		R = (colour >> 16) & 0xff,
-		G = (colour >>  8) & 0xff,
-		B = (colour >>  0) & 0xff,
+		G = (colour >>	8) & 0xff,
+		B = (colour >>	0) & 0xff,
 		Alpha = PV_Linear(255, 0),
 		ForceY = PV_Gravity(100),
 		CollisionVertex = 0
@@ -177,7 +177,7 @@ private func DrawWaterSource()
 	}
 }
 
-private func DoWaterdrop()
+public func OnRainDropCreated()
 {
 	if (Random(9)) return;
 	Sound("Liquids::Waterdrop*");
--- a/planet/Objects.ocd/Environment.ocd/Stalactite.ocd/Script.c
+++ b/planet/Objects.ocd/Environment.ocd/Stalactite.ocd/Script.c
@@ -21,7 +21,7 @@ public func Construction()
 		sibling = TransformBone(bone, transformation, 1, Anim_Const(1000), sibling);
 	}
 	
-	AddRainDropEffect(nil, 100, "Water", 2);
+	AddRainDropEffect(nil, RandomX(80, 120), "Water", RandomX(1, 5));
 }
 
 private func Hit()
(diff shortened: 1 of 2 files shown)


planet/Objects.ocd/Environment.ocd/Stalactite.ocd/Script.c
```diff
@@ -31,8 +31,8 @@ private func Hit()
 	{
 		Size = PV_KeyFrames(0, 0, 0, 100, PV_Random(3, 5), 1000, 3),
 		R = (colour >> 16) & 0xff,
-		G = (colour >>  8) & 0xff,
-		B = (colour >>  0) & 0xff,
+		G = (colour >>	8) & 0xff,
+		B = (colour >>	0) & 0xff,
```
walachey: In the future, please refrain from mixing format fixes for lines you usually wouldn't touch into a commit. That can really mess up merges big time - I remember when someone randomly renamed a variable and added parameter types in an unrelated commit and messed up my Controls branch :[
If you want to change formatting, do a separate commit (and in this case using tabs for non-indention formatting actually conflicts with our style guidelines)

planet/Objects.ocd/System.ocg/Effects.c
```diff
@@ -79,3 +79,79 @@ global func ExplosionEffect(int level, int x, int y, int smoothness, bool silent
...
+			this.particle_cache.particle.Size = 5;
+		}
+		
+		if (this.material == "Lava" || this.material == "DuroLava")
+			particle_name = "RaindropLava";
+		
+		else
+			particle_name = "Raindrop";
+		
+		for (var i = 0; i < this.strength; i++)
```
walachey: This loop is probably unnecessary - CreateParticle can take random value providers (PV_Random) for the position and velocity. So you can replace the loop with `CreateParticle(..., PV_Random(-wdt, wdt), PV_Random(-hgt, hgt), PV_Random(-5, 5), PV_Random(10, 30), ...)` and set the amount parameter to `this.strength`. That should also be a lot faster, because now the particle proplist has to be parsed for every new particle - batch creation is a lot faster there!

planet/Objects.ocd/System.ocg/Effects.c
```diff
@@ -79,3 +79,79 @@ global func ExplosionEffect(int level, int x, int y, int smoothness, bool silent
...
+		{
+		    var x = RandomX(-wdt, wdt);
+		    var y = RandomX(-hgt, hgt);
+		    var xdir = RandomX(-5, -5);
+		    var ydir = RandomX(10, 30);
+			this.Target->CreateParticle(
```
walachey: Indentation is messed up - you mix four-space-indented code with tabs-indented code

planet/Objects.ocd/System.ocg/Effects.c
walachey: PS: in this case, your loop would also make the raindrop sound play a lot more often compared to before (because the `if (!Random(9))` in the stalactite is evaluated multiple times per frame - given that `this.strength` is `>1`

planet/Objects.ocd/Environment.ocd/Stalactite.ocd/Script.c
```diff
@@ -21,7 +21,7 @@ public func Construction()
 		sibling = TransformBone(bone, transformation, 1, Anim_Const(1000), sibling);
 	}
 	
-	AddTimer("DoWaterdrop", RandomX(36 * 10, 36 * 40));
+	AddRainDropEffect(nil, 100, "Water", 2);
```
walachey: Because the raindrop interval is not randomized, all stalactites in the game will drop at the same time (and also play the sound at the same time), which might be veeeery weird ingame. Also you drop & sound around nine times more often than before (on average)

planet/Objects.ocd/System.ocg/Effects.c
walachey: PPS: using value-provider functions also leads to nicer looking effects. Because `RandomX(-1, 1)` can be either {-1, 0, +1}, but `PV_Random(-1, +1) `can also be 0.232154 (cause it's not sync-relevant).

planet/Objects.ocd/System.ocg/Effects.c
Fulgen301: What. How did that happen? ô.o

planet/Objects.ocd/System.ocg/Effects.c
Fulgen301: Okay, will be fixed, as well as the other remarks.

planet/Objects.ocd/Environment.ocd/Stalactite.ocd/Script.c
Fulgen301: Hm. I didn't even touch that. Perhaps Notepad++ messed something up.

walachey:
Looks good now. Is this ready to be merged? Then I'd squash the three commits into the first and pick it into the OC repos

walachey:
Merged

PR #48, opened 2017-06-10
Remove StringToInteger function

Fulgen301:
The engine-internal ParseInt function uses the same parsing rules as the engine constant parser and should be preferred.


lluchs:
I think it's currently only used by @MDT-Maikel's tower: https://github.com/MDT-Maikel/OCTower/blob/69868cdde3440825c18f001601fea3c967d2f9da/TowerObjects.ocd/System.ocg/Saving.c#L24

He'll want to fix the tower before merging this.

PR #51, opened 2017-07-11
Share time information (frameCounter and realtime) with landscape shaders

ckanibal:
ezgif-4-042f1b8f99

It uses a int32[2]-vector for the realtime because ARB_gpu_shader_fp64 is not very well supported before OpenGL 4.0 on some Intel GPUs.

### DIFF
--- a/src/graphics/C4Shader.h
+++ b/src/graphics/C4Shader.h
@@ -210,6 +210,34 @@ public:
 		if (pShader->HaveUniform(iUniform))
 			glUniform1i(pShader->GetUniform(iUniform), iX);
 	}
+	void SetUniform2i(int iUniform, int iX, int iY) const {
+		if (pShader->HaveUniform(iUniform))
+			glUniform2i(pShader->GetUniform(iUniform), iX, iY);
+	}
+	void SetUniform3i(int iUniform, int iX, int iY, int iZ) const {
+		if (pShader->HaveUniform(iUniform))
+			glUniform3i(pShader->GetUniform(iUniform), iX, iY, iZ);
+	}
+	void SetUniform4i(int iUniform, int iX, int iY, int iZ, int iW) const {
+		if (pShader->HaveUniform(iUniform))
+			glUniform4i(pShader->GetUniform(iUniform), iX, iY, iZ, iW);
+	}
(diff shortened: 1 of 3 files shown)


ckanibal:
I changed the epoch so that it now counts the milliseconds since the shader initialization.
This way it only needs one int32 - a possible overflow might occur after 596.5 hours which I think is acceptable.

kpone33:
I agree with that. I doubt there will ever be someone playing for 24 days without any interuption.

src/landscape/C4LandscapeRender.cpp
```diff
@@ -71,6 +71,9 @@ bool C4LandscapeRenderGL::Init(int32_t iWidth, int32_t iHeight, C4TextureMap *pT
 	this->iHeight = iHeight;
 	this->pTexs = pTexs;
 
+	// Start timer
+	this->TimerStart = std::chrono::system_clock::now();
```
isilkor: `std::chrono::system_clock` isn't guaranteed to be monotonic; use `std::chrono::steady_clock` instead.

src/graphics/C4Shader.h
```diff
@@ -210,6 +210,10 @@ class C4ShaderCall
 		if (pShader->HaveUniform(iUniform))
 			glUniform1i(pShader->GetUniform(iUniform), iX);
 	}
+	void SetUniform2i(int iUniform, int iX, int iY) const {
+		if (pShader->HaveUniform(iUniform))
+			glUniform2i(pShader->GetUniform(iUniform), iX, iY);
+	}
```
isilkor: Do you still need this function?

Fulgen301:
@kpone33 What about savegames?

isilkor:
Savegames are unaffected by this because the elapsed time isn't stored 
anywhere (it's time since current rendering started).

src/graphics/C4Shader.h
ckanibal: The function is not used currently, but the same holds true for SetUniform1iv and others. I guess it's okay to leave it there right now 🤞

isilkor:
Can you split it out of the timer commit then (and add the remaining 
missing functions, if any)?

PR #53, opened 2017-08-26
Modernize loader loading code (#1748)

ckanibal:
This should fix #1748. 
Note: It seems that the engine is not searching for loaders in the currently opened scenario group.

### DIFF
--- a/src/gui/C4LoaderScreen.cpp
+++ b/src/gui/C4LoaderScreen.cpp
@@ -25,6 +25,11 @@
 #include "lib/C4LogBuf.h"
 #include "lib/C4Random.h"
 
+// allow combination of loader flags
+inline C4LoaderScreen::Flag operator|(C4LoaderScreen::Flag a, C4LoaderScreen::Flag b) {
+	return static_cast<C4LoaderScreen::Flag>(static_cast<int>(a) | static_cast<int>(b));
+}
+
 C4LoaderScreen::C4LoaderScreen()
 {
 	// zero fields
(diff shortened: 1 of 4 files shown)


src/gui/C4LoaderScreen.cpp
```diff
@@ -113,29 +104,25 @@ void C4LoaderScreen::SetBlackScreen(bool fIsBlack)
...
-		if (!UnsyncedRandom(++iLoaderCount))
-		{
-			// copy group and path
-			*ppDestGrp=&rFromGrp;
-			SCopy(Filename, szDstName, _MAX_PATH);
+		// potential candidate - check file extension
+		std::string extension{ GetExtension(filename) };
+		if (extension == "png" || extension == "jpg" || extension == "jpeg" || extension == "bmp") {
+			auto loader = std::pair<C4Group*, std::string>{ &rFromGrp, std::string(filename) };
+			loaders.emplace(loader);
```
lluchs: Can't you do this directly as `loaders.emplace(&rFromGrp, std::string(filename));`?

src/gui/C4LoaderScreen.h
```diff
@@ -13,29 +13,45 @@
...
 #ifndef INC_C4LoaderScreen
 #define INC_C4LoaderScreen
 
 #include "graphics/C4FacetEx.h"
 
+
 class C4LoaderScreen
 {
 public:
+	enum Flag {
```
lluchs: Should be `enum class` as you're always using `Flag::FOO` anyways.

src/gui/C4LoaderScreen.h
```diff
@@ -13,29 +13,45 @@
...
+		ALL = 0xFF
+	};
+
+protected:
 	C4FacetSurface fctBackground; // background image
 	char *szInfo;      // info text to be drawn on loader screen
 	bool fBlackScreen; // if set, a black screen is drawn instead of a loader
 
+	std::map<C4Group*, const std::string> loaders;
+	void SeekLoaderScreens(C4Group &rFromGrp, const std::string wildcard);
```
isilkor: Do you really want a const copy here, or should wildcard be a const reference?

src/gui/C4LoaderScreen.h
ckanibal: This would reqire C4LoaderScreen Flags to be strongly typed - requiring all the bitwise-operators that one has learned to love needed to be reimplemented. I guess it's less effort this way.

lluchs: Ah. It looked like you did that anyways.

ckanibal: Yup, I did. Then I noticed in my latest changes that it isn't needed this way, so I removed the superfluous methods.

PR #54, opened 2017-10-05
 Replace C4Network2HTTPClient with the curl-based C4HTTPClient

lluchs:
The old HTTP client is pretty bad: it doesn't support TLS, it only speaks HTTP 1.0, it needs headers set in a certain way. By replacing it with curl, we get proper HTTP and HTTPS support.

I chose curl as that's a widely-used HTTP client library. It is already installed on most Linux systems and (as far as I can tell) is easy to compile for Windows.

There's some Windows-specific code in the PR which I couldn't test.

`CURLOPT_VERBOSE` is still enabled for debugging, I'm going to remove that before merging.


walachey:
I could build it on windows without problems - curl seems to build and work fine (with CMake).
However, we likely want a way to have our CMake find curl automatically. I had to set the include path and library path manually.

However, something is off about the scheduler(?). On the first try, the network view worked fine. On successive trials, I always got a timeout. It crashes when I hold down F5 and it also crashes eventually when running in the background.
On my mods dialogue, I had a similar crash. It resulted from adding callbacks to the scheduler from the wrong thread with the wrong timing - I can probably find the commit, if it would help.
Do we want to use one scheduler/one thread for things like that? Spawning threads is likely the least delaying operation about a curl request, so imo we could as well just have each requester be a separate thread. @isilkor?

isilkor:
David Dormagen wrote on 06.10.2017 12:32:
> I could build it on windows without problems - curl seems to build and 
> work fine (with CMake).
> However, we likely want a way to have our CMake find curl automatically. 
> I had to set the include path and library path manually.
With the amount of #ifdef blocks we're already having to use in the HTTP 
client, I'm almost tempted to look into just using WinHTTP on Windows.

> On my mods dialogue, I had a similar crash. It resulted from adding 
> callbacks to the scheduler from the wrong thread with the wrong timing - 
Ugh. This is something the scheduler should really deal with; our 
threading code should be, well, thread-safe.

> Do we want to use one scheduler/one thread for things like that? 
> Spawning threads is likely the least delaying operation about a curl 
> request, so imo we could as well just have each requester be a separate 
> thread.We could, but I've no idea about the thread safety of libcurl.

lluchs:
>With the amount of #ifdef blocks we're already having to use in the HTTP client, I'm almost tempted to look into just using WinHTTP on Windows.

Why does the Windows StdScheduler need a completely different API anyways? Couldn't it just use GetFDs() with WSAPoll()? That would probably allow some code reduction in the NetIO classes as well.

>>Do we want to use one scheduler/one thread for things like that? Spawning threads is likely the least delaying operation about a curl request, so imo we could as well just have each requester be a separate thread. 
>We could, but I've no idea about the thread safety of libcurl.

I believe curl works fine with threads as long as you use one handle from one thread only. It's better to reuse a curl handle between requests to the same server though to benefit from keepalive. This would probably make for a somewhat annoying-to-use interface with threads.

isilkor:
> Why does the Windows StdScheduler need a completely different API 
> anyways? Couldn't it just use GetFDs() with WSAPoll()?

IIRC WSAPoll only works for Winsock stuff, so you can't use it to wait 
for events or other handles.

lluchs:
I updated the Windows code (still using the same approach with the scheduler instead of a separate thread). I can't test it from the UI, but I opened a lobby with openclonk-server and let it send updates to the masterserver for a while, which worked fine. @walachey can you test this again, please?

antony-jr:
@lluchs Since Qt is a dependency , why not use Qt Network Module. It has thread-safety and everything you need. It is also easy to use and well documented.

gitMarky:
Is this still relevant? I guess so, but still wanted to check. What is preventing us to merge, the windows-specific code?

lluchs:
Yes, no point in merging if it doesn't work on Windows. We should have something like this though before releasing the mod downloading. I'm considering the suggestion above to use Qt, might be the easiest way.

src/network/C4HTTP.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+bool C4HTTPClient::Execute(int iMaxTime, pollfd *readyfd)
+{
+	int running;
+#ifdef STDSCHEDULER_USE_EVENTS
+	// On Windows, StdScheduler doesn't inform us about which fd triggered the
+	// event, so we have to check manually.
+	if (WaitForSingleObject(Event, 0) == WAIT_OBJECT_0)
+	{
+		for (const auto& kv : sockets)
```
Fulgen301, line 67: This is going to crash on Windows as the socket callback called by `curl_multi_socket_action` may erase the socket out of the map, and changing the map in the body of a ranged for loop isn't allowed.

Fulgen301 (changes requested):

PR #55, opened 2017-10-21
added docs for OpenClonk master

marsmoon:
This allows to make updated docs for the latest OpenClonk version from master branch while keeping separate docs for the last stable version. It will be possible to keep the documentation up-to-date between stable releases.

oc-docs


lluchs:
Thanks for the pull request! I agree that online documentation for the stable version is currently missing. I'm not sure whether this is the best way to achieve this. Someone checking out the repository is usually only interested in documentation for that version.

@westnordost I believe you're maintaining the online documentation. Do you think you could provide a static copy of the stable documentation, or would you prefer a solution like the one in the PR here?

Nevertheless, some comments on the code:

 - You accidentally committed a swap file in `docs/sdk-latest/.index.xml.swp`
 - I think we should keep the existing documentation files as "latest" and add a copy of the old documentation from the 7.0 tag.
 - Translations should be generated for both documentation variants.

marsmoon:
> Someone checking out the repository is usually only interested in documentation for that version.
That is true, but I think it would be much better if both versions could be available online.

> I think we should keep the existing documentation files as "latest" and add a copy of the old documentation from the 7.0 tag.
I'm a little confused right now. Is existing documentation for the latest version or is it for stable? I assumed it was for stable, because some things are outdated and it doesn't say anywhere in the docs what version they are for. If the current one is for stable will that not be confusing to most users? I think most users use stable version (especially if they are new to the game), so stable should be the default documentation that they see. Stable is probably the only version available in GNU/Linux repositories.

> Translations should be generated for both documentation variants
I don't know how to do that. I will need some help.

westnordost:
@lluchs Neither. I don't want to manually maintain a copy of the documentation and I think the repository is the wrong place to put a copied "last-stable version" documentation. We don't put our releases in our repository either.

Instead, this should be done through the continuous integration scripts from @aburgm 

In my opinion, we do not need a "nightly" documentation next to a "last stable" documentation. I am not sure if the current documentation reflects the current state of repository or the last stable right now. The official documentation on the OpenClonk website should always be only the "last stable" documentation.

Anyhow, if the two really have become so different, maybe it is high time to ship out another release instead of puttin more effort into on how to better support both the stable and the repo version. I mean, it has been almost two years now, what are you (=the active developers) afraid of?

marsmoon:
Those two versions really are different. Someone pointed out that variable Components has been removed from DefCore (http://docs.openclonk.org/en/sdk/definition/defcore.html) and made accessible inside object's script instead. Looks like this change was made around 30 april 2016 (https://github.com/openclonk/openclonk/commit/1df322b7ffd59352a09d74c2af290469d33bbeb2#diff-f0c9194728ea71ce17188496c3e51980) and yet both information on the website (http://docs.openclonk.org/en/sdk/definition/script.html) and in the repo (https://github.com/openclonk/openclonk/blob/master/docs/sdk/definition/script.xml) are still outdated. This is just one example, there might be more. I don't know what else has changed.

I would like quicker releases of OpenClock too, but let's assume that 8.0 is released tomorrow. How much time will it take for another release (v 9.0)? Will it take a year or two years for another version to come out? If future versions were always released every month I would probably agree, but I think that if we keep doing things like we do right now, the docs will never be good.

The only problem with having two different versions that I can see is: adding something missing that is relevant to both versions.

walachey:
>he only serious problem that having two different versions could cause is when it's needed to add something missing that is relevant to both versions. Then it would be required to make changes in two different places instead of one, which is not great. Maybe there is some way to solve this issue?

The 'correct' solution would probably to use git for that. And we do that already for the releases! We tag a version as e.g. "7.0" and then continue working. If we have changes that we need both in a bugfix release for 7.0 and in the next release, we put them on master AND add cherry-pick them onto a branch based off of the "7.0" tagged version.
And the CI scripts can both release from tags and from master - so we can ship a bugfix release for an old version without maintaining two /places/ (just two points in history, using git).

And that's probably also how the docs on the website should be packed: either from 'master' or from the revision of the last release (or a branch on top of that). Then we don't have the problem with the languages.

But thanks for bringing that to attention. The current situation really is *bad*... Currently, the nightly documentation is on the website and we don't keep the last release's docs around.
@aburgm, what would be the required effort to have both versions around? @isilkor, comments?

walachey:
No changes since the last comment. The correct solution would still be using CI for this instead of maintaining a "version" as a "language" manually. Closing the PR.
Thank you for the effort you put into this, marsmoon!

lluchs:
The documentation is now built on [sr.ht](https://builds.sr.ht/~luchs/openclonk/docs.yml), but still only for master. I'll look into keeping around release versions when redoing the rest of the release infrastructure.

PR #57, opened 2017-10-29
Replaced a comment with more accurate one

marsmoon:
It was easy to forget about updating the old comment (https://github.com/openclonk/openclonk/commit/5c45d7a08e1f9a9dc691248cb781b6119853ff35) after changing the
code below it, which would make that comment confusing and irrelevant. It
wasn't accurate, because those weren't the only items that could be used
to make Tele Glove. It also had a typo in word diamond.


lluchs:
Seems to be obsolete now.

PR #59, opened 2017-10-30
Docs: added an example for appendto

marsmoon:
Example usage of #appendto.


lluchs:
Thanks! I simplified the example a bit to show `inherited` usage as well.

PR #61, opened 2017-12-13
Lightning: Fix Find_Exclude error

Fulgen301:


walachey:
Thanks Fulgen301! I pushed a similar solution now where the check happens once in LaunchLightning and not every frame in the timer call.
(and where the commit message explains something about the issue)

Issue #64, opened 2018-02-06
openclonk-0.8 linking fails on Fedora 27


martinkg:
Hi,

compiling openclonk [1] fails with the following error message [2]:

[  9%] Linking CXX static library liblibmisc.a
/usr/bin/cmake -P CMakeFiles/libmisc.dir/cmake_clean_target.cmake
/usr/bin/cmake -E cmake_link_script CMakeFiles/libmisc.dir/link.txt --verbose=1
[...]

PR #65, opened 2018-02-08
Fix format-security GCC error (GH-64)

akien-mga:
Like Fedora, Mageia builds all its packages with `-Werror=format-security`, so even if this code branch is actually never used as mentioned in https://github.com/openclonk/openclonk/issues/64#issuecomment-363591638, at least it builds :)

### DIFF
--- a/src/script/C4AulCompiler.cpp
+++ b/src/script/C4AulCompiler.cpp
@@ -74,7 +74,7 @@ static void Warn(const C4ScriptHost *target_host, const C4ScriptHost *host, cons
 		return;
 	}
 	const char *msg = C4AulWarningMessages[static_cast<size_t>(warning)];
-	std::string message = sizeof...(T) > 0 ? strprintf(msg, std::forward<T>(args)...) : msg;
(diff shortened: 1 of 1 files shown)


lluchs:
Although your commit fixes the warning, it's incorrect and breaks the code using the function.

There are two cases here:

1. The function is called as `Error([...], "this is msg");`. The template expands to:

 ```cpp
std::string message = 0 > 0 ? strprintf("this is msg") : "this is msg";
```

You'll see that the middle part is dead code (0 is never greater than 0). GCC still issues the `-Wformat-security` warning incorrectly.

2. The function is called as `Error([...], "this is msg with %d formatting", 42);`. The template expands to:

 ```cpp
std::string message = 1 > 0 ? strprintf("this is msg with %d formatting", 42) : "this is msg with %d formatting";
```

Now `strprintf` is always called and doesn't cause any warnings. It will print "this is msg with 42 formatting". With your change, it will not do formatting in msg and will instead print "this is msg with %d formatting".

lluchs:
If you really don't want to disable `-Werror=format-security` globally, we may merge a patch that uses `#pragma`s to disable the warnings for that function only, if you want to try doing that :)

PR #67, opened 2018-02-12
Update openclonk.desktop

ronnystandtke:
- added a German translation of the GenericName field
- left out the "multiplayer" part, because OpenClonk is also a very fine single player game but this would make the German translation much too long...


isilkor:
Just merged this. Sorry for the wait, and thank you for the contribution!

PR #71, opened 2018-04-10
Stable 8 for upstream

RomainNaour:
Hi,

Following the discussion last summer [1].
Here are some patches to allow building OpenClonk with Buildroot.

The last patch (add C4GROUP_TOOL_ONLY) is not really nice.
It would be great to be able to just compile c4goup tool without add a lot of complexity in the CMake build system.

This series is based on the stable-8 branch.
I can rebase against master if needed.

[1] https://forum.openclonk.org/topic_show.pl?tid=3391

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -990,7 +990,7 @@ include_directories(SYSTEM
 ############################################################################
 add_definitions(-DHAVE_CONFIG_H)
 
-add_library(libmisc
+add_library(libmisc STATIC
 src/C4Include.cpp
 src/c4group/C4Group.cpp
 src/c4group/C4Group.h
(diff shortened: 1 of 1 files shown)


ckanibal:
I'm a big fan of the buildroot project - I've used it before to assemble a firmware image for a DSL router of mine.

I wonder what kind of embedded systems would be capable of running OpenClonk? There's this somewhat ambitious OpenGL 3.2 barrier that you would need to overcome first. Are you targeting x86?

RomainNaour:
Hi @jcaesar, 
Sure, I'll squash the two first patches.
I'm also not really happy with the last patch. Maybe we should create a sub project for C4GROUP tool ?

@ckanibal : Thanks, I'm adding some cool video games to Buildroot for the fun. Note, there are already Supertux, Supertuxkart, Minetest, Lugaru and other in Buildroot.
Indeed I'm targeting an x86 but this allow me to test new mesa version (currently mesa 18.0).
Also, since piglit tool is not really fun ;-), I'm adding these games.

PR #84, opened 2019-01-26
Add Flintbrawl by Zapper

lluchs:
I changed the script to use the new multi-round melee goal and wrote a description. Needs confirmation by @walachey because it's his scenario.

Possible todos:

 - [ ] Move the new flints to Objects.ocd?
 - [x] Convert sounds to vorbis
 - [x] Move sounds to Sound.ocg?
 - [x] I think there was some issue with the team mode. Check that it works or debug the problem.
 - [x] Icon is still the snow icon
 - [ ] squash some of the commits


Fulgen301:
>Move the new flints to Objects.ocd?
Do they add a valuable experience to the default game?

lluchs:
>Do they add a valuable experience to the default game?

I think the question is rather: Are they useful to other scenario authors? Would people expect them to appear in the editor?

walachey:
I am all for moving the scenario into the default pack. Converting the sounds to .ogg seems like a good idea (and probably putting them into the default Sound.ocg as well - the glass sound is pretty generic)

I am not sure whether the flints fit into the original package. But I am pretty convinced that we need more stuff there, sooo maybe (e.g. the chippy flint).

>I think there was some issue with the team mode. Check that it works or debug the problem.
There once was a desync, but (I think?) I had fixed that already. Is that what you meant? If so, that was reproducible every round, so it should be easy to check.

Anything you want me to do or need feedback on?

lluchs:
>There once was a desync, but (I think?) I had fixed that already. Is that what you meant? If so, that was reproducible every round, so it should be easy to check.

Okay, I'll check that. I don't think I've ever played the scenario with teams, so I wasn't sure.

>Anything you want me to do or need feedback on?

You did the graphics for the new flints, right? So that the licensing is clear. Otherwise, I think everything else is clear.

lluchs:
We tried the scenario with teams today, and we got the desync. Maybe I didn't have the latest version? Otherwise I will look into it.

lluchs:
So the issue with the desync is that there is disagreement about `GetTeamCount()` when setting the teams to surprise random. On the host, `GetTeamCount()` returns 2 (the actual number of teams), but on the clients, it seems to return the number of teams before changing to surprise random (e.g., 3 if there were three players).

walachey:
I checked my local version and have no GetTeamCount call in the scenario script anymore, so I guess I had already fixed that. In case I also did other changes, I'd like to either give you my local version to check for differences or do that myself. What would you prefer and how much of the directory structure etc. did you change from the original?

walachey:
> Just give me your latest version. The only file with significant changes by me is Script.c, I think, so it should be fairly easy to merge.

Here you go http://userpage.fu-berlin.de/dormagen/clonk/FlintBrawl.ocs.zip
Thanks for taking this on

Fulgen301:
>I think the question is rather: Are they useful to other scenario authors? Would people expect them to appear in the editor?

Are they so useful that they have to be loaded with every single scenario? Because that's what happens when they get put in Objects.ocd, and while they're neat in this specific scenario, I doubt they fit into other scenarios, so maybe they should stay in the scenario as local definitions.
Also, if a scenario developer needs them, he can specify the definition in Scenario.txt!

lluchs:
>Also, if a scenario developer needs them, he can specify the definition in Scenario.txt!

Loading definitions from other scenarios is not intuitive at all and has a high risk of breaking later on if the scenario changes.

lluchs:
Good, I updated it now. Removing the `GetTeamCount()` call isn't necessary anymore with my engine fix on master, but your updated version still seemed cleaner. Should be fine for merging now.

PR #87, opened 2019-03-05
Refactor

gitMarky:
First humble changes in the movement code. No idea how to proceed from there, just review it and lets see what comes of it.

I also want to get rid of the nested if()s and replace them with && operators.

### DIFF
--- a/src/object/C4Movement.cpp
+++ b/src/object/C4Movement.cpp
@@ -57,7 +57,7 @@ void ApplyFriction(C4Real &tval, int32_t percent)
 	{
 		tval -= ffric;
 	}
-	else if (tval <- ffric)
+	else if (tval < -ffric)
 	{
 		tval += ffric;
 	}
--- a/src/object/C4Movement.cpp
+++ b/src/object/C4Movement.cpp
@@ -56,14 +56,15 @@ void ApplyFriction(C4Real &tval, int32_t percent)
 	if (tval > +ffric)
 	{
 		tval -= ffric;
-		return;
 	}
-	if (tval <- ffric)
+	else if (tval <- ffric)
 	{
 		tval += ffric;
-		return;
 	}
-	tval = 0;
+	else
+	{
+		tval = 0;
+	}
 }
 
 // Compares all Shape.VtxContactCNAT[] CNAT flags to search flag.
--- a/src/object/C4Movement.cpp
+++ b/src/object/C4Movement.cpp
@@ -28,41 +28,42 @@
 
 /* Some physical constants */
 
-const C4Real FRedirect=C4REAL100(50);
-const C4Real FFriction=C4REAL100(30);
-const C4Real FixFullCircle=itofix(360),FixHalfCircle=FixFullCircle/2;
-const C4Real FloatFriction=C4REAL100(2);
-const C4Real RotateAccel=C4REAL100(20);
-const C4Real HitSpeed1=C4REAL100(150); // Hit Event
-const C4Real HitSpeed2=itofix(2); // Cross Check Hit
-const C4Real HitSpeed3=itofix(6); // Scale disable, kneel
-const C4Real HitSpeed4=itofix(8); // Flat
-const C4Real DefaultGravAccel=C4REAL100(20);
+const C4Real FRedirect = C4REAL100(50);
+const C4Real FFriction = C4REAL100(30);
+const C4Real FixFullCircle = itofix(360);
+const C4Real FixHalfCircle = FixFullCircle / 2;
+const C4Real FloatFriction = C4REAL100(2);
+const C4Real RotateAccel = C4REAL100(20);
+const C4Real HitSpeed1 = C4REAL100(150); // Hit Event
+const C4Real HitSpeed2 = itofix(2); // Cross Check Hit
+const C4Real HitSpeed3 = itofix(6); // Scale disable, kneel
+const C4Real HitSpeed4 = itofix(8); // Flat
+const C4Real DefaultGravAccel = C4REAL100(20);
 
 /* Some helper functions */
 
 void RedirectForce(C4Real &from, C4Real &to, int32_t tdir)
 {
 	C4Real fred;
-	fred=std::min(Abs(from), FRedirect);
-	from-=fred*Sign(from);
-	to+=fred*tdir;
+	fred = std::min(Abs(from), FRedirect);
+	from -= fred * Sign(from);
+	to += fred * tdir;
 }
 
 void ApplyFriction(C4Real &tval, int32_t percent)
 {
-	C4Real ffric=FFriction*percent/100;
-	if (tval>+ffric)
+	C4Real ffric = FFriction * percent / 100;
+	if (tval > +ffric)
 	{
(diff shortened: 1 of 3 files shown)


src/object/C4Movement.cpp
```diff
@@ -136,17 +159,25 @@ void C4Object::StopAndContact(C4Real & ctco, C4Real limit, C4Real & speed, int32
...
-	t_contact=Shape.ContactCNAT;
+	t_contact = Shape.ContactCNAT;
 
 	// Contact script call for the first contacted cnat
 	if (Shape.ContactCNAT)
-		for (int32_t ccnat=0; ccnat<4; ccnat++) // Left, right, top bottom
+	{
+		for (int32_t ccnat = 0; ccnat < 4; ccnat++) // Left, right, top bottom
+		{
 			if (Shape.ContactCNAT & (1<<ccnat))
```
isilkor: Since you're adding whitespace everywhere else, the shifts here should get some too

src/object/C4Movement.cpp
```diff
@@ -186,50 +227,74 @@ void C4Object::VerticalBounds(C4Real &ctcoy)
...
 {
-	int32_t iContact=0;
-	bool fAnyContact=false; int iContacts = 0;
-	BYTE fTurned=0,fRedirectYR=0,fNoAttach=0;
+	int32_t iContact = 0;
+	bool fAnyContact = false;
+	int iContacts = 0;
+	BYTE fTurned = 0,
+		 fRedirectYR = 0,
+		 fNoAttach = 0;
```
isilkor: Can you turn these into individual declarations, per http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-name-one?

src/object/C4Movement.cpp
```diff
@@ -186,50 +227,74 @@ void C4Object::VerticalBounds(C4Real &ctcoy)
...
+				}
+				::Landscape.DigFree(ctcox, ctcoy - 1, rad, this);
 			}
 		}
+	}
 
 	// store previous movement and ocf
-	C4Real oldxdir(xdir), oldydir(ydir);
+	C4Real oldxdir(xdir),
+		   oldydir(ydir);
```
isilkor: Same (individual declarations)

src/object/C4Movement.cpp
```diff
@@ -275,31 +341,36 @@ void C4Object::DoMovement()
...
 				else
 				{
 					// living things are always capable of keeping their rotation
-					if (OCF & OCF_Rotate) if (iContact==1) if (!Alive)
-							{
-								RedirectForce(ydir,rdir,-ContactVtxWeight(this));
-								fRedirectYR=1;
-							}
-					ydir=0;
+					if (OCF & OCF_Rotate) if (iContact == 1) if (!Alive)
```
isilkor:
```suggestion
					if ((OCF & OCF_Rotate) && iContact == 1 && !Alive)
```

src/object/C4Movement.cpp
```diff
@@ -311,45 +382,57 @@ void C4Object::DoMovement()
 		do
 		{
 			// Set next step target
-			int step_x = 0, step_y = 0;
+			int step_x = 0,
+				step_y = 0;
```
isilkor: Declarations again.

src/object/C4Movement.cpp
```diff
@@ -392,59 +483,80 @@ void C4Object::DoMovement()
...
 				// Undo step and abort movement
-				Shape=lshape;
+				Shape = lshape;
 				target_r = fix_r = lcobjr;
 				// last UpdateShape-call might have changed sector lists!
 				UpdatePos();
 				// Redirect to GetY()
-				if (iContact==1) if (!fRedirectYR)
-						RedirectForce(rdir,ydir,-1);
+				if (iContact == 1) if (!fRedirectYR)
```
isilkor:
```suggestion
				if (iContact == 1 && !fRedirectYR)
```

src/object/C4Object.h
```diff
@@ -327,6 +327,7 @@ class C4Object: public C4PropListNumbered
 protected:
 	void SideBounds(C4Real &ctcox);       // apply bounds at side; regarding bourder bound and pLayer
 	void VerticalBounds(C4Real &ctcoy);   // apply bounds at top and bottom; regarding border bound and pLayer
+	void RemoveSolidMask(bool fBackupAttachment); // Remove solid mask data, if existing
```
isilkor: This can be private.

src/object/C4Movement.cpp
```diff
@@ -658,11 +841,17 @@ bool SimFlightHitsLiquid(C4Real fcx, C4Real fcy, C4Real xdir, C4Real ydir)
 	// Start in water?
 	int temp;
 	if (DensityLiquid(GBackDensity(fixtoi(fcx), fixtoi(fcy))))
-		if (!SimFlight(fcx, fcy, xdir, ydir, 0, C4M_Liquid - 1, temp=10))
+	{
+		if (!SimFlight(fcx, fcy, xdir, ydir, 0, C4M_Liquid - 1, temp = 10))
```
isilkor: With the introduction of that block, `temp` can be moved inside and initialized there instead of that horrible assignment during parameter passing. (Same below.)

src/object/C4Movement.cpp
gitMarky: Ok

src/object/C4Movement.cpp
gitMarky: Yes, I prefer that, but was not sure what is better. Good to know!

src/object/C4Movement.cpp
gitMarky:
```suggestion
					if ((OCF & OCF_Rotate) && !Alive && iContact == 1)
```
Did not really understand how suggestion works. Anyway, I think that the order can be changed, maybe even have !Alive first.

gitMarky: What about equality checks inverted? 1 == iContact

isilkor: That reads like Yoda and isn't really useful because compilers like to complain about accidental assignments.

src/object/C4Movement.cpp
gitMarky: This is resolved now, right?

src/object/C4Movement.cpp
gitMarky: Funny though, because we had several people at work who thought that this is super safe for avoiding accidental assignments (well, it is) and the thing to always do.

isilkor:
It is safe(r), but clang warns about it by default, gcc does it as part 
of -Wall (or explicit by -Wparentheses, which works for clang too), and 
for MSVC you need to enable it explicitly with /we4706.

src/object/C4Movement.cpp
isilkor: It is safe(r), but clang warns about it by default, gcc does it as part of `-Wall` (or explicit by `-Wparentheses`, which works for clang too), and for MSVC you need to enable it explicitly with `/we4706`.

PR #88, opened 2019-03-06
Refactor movement: Further cleanup, added new function

gitMarky:
This is still not the end, but it is a small step forward :)

### DIFF
--- a/src/object/C4Movement.cpp
+++ b/src/object/C4Movement.cpp
@@ -156,10 +156,10 @@ void C4Object::StopAndContact(C4Real & ctco, C4Real limit, C4Real & speed, int32
 	Contact(cnat);
 }
 
-int32_t C4Object::ContactCheck(int32_t iAtX, int32_t iAtY, uint32_t *border_hack_contacts, bool collide_halfvehic)
+int32_t C4Object::ContactCheck(int32_t at_x, int32_t at_y, uint32_t *border_hack_contacts, bool collide_halfvehic)
 {
 	// Check shape contact at given position
-	Shape.ContactCheck(iAtX, iAtY, border_hack_contacts, collide_halfvehic);
+	Shape.ContactCheck(at_x, at_y, border_hack_contacts, collide_halfvehic);
 
 	// Store shape contact values in object t_contact
 	t_contact = Shape.ContactCNAT;
@@ -184,7 +184,7 @@ int32_t C4Object::ContactCheck(int32_t iAtX, int32_t iAtY, uint32_t *border_hack
 }
 
 // Stop the object and do contact calls if it collides with the border
-void C4Object::SideBounds(C4Real &ctcox)
+void C4Object::SideBounds(C4Real &target_x)
 {
 	// Layer bounds
 	if (Layer && (Layer->GetPropertyInt(P_BorderBound) & C4D_Border_Layer))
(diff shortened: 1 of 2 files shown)


src/object/C4Movement.cpp
```diff
@@ -44,10 +44,10 @@ const C4Real DefaultGravAccel = C4REAL100(20);
 
 void RedirectForce(C4Real &from, C4Real &to, int32_t tdir)
 {
-	C4Real fred;
-	fred = std::min(Abs(from), FRedirect);
-	from -= fred * Sign(from);
-	to += fred * tdir;
+	C4Real force_redirected;
+	force_redirected = std::min(Abs(from), FRedirect);
```
isilkor: You can combine the variable definition and assignment here.

src/object/C4Movement.cpp
```diff
@@ -184,122 +182,99 @@ int32_t C4Object::ContactCheck(int32_t iAtX, int32_t iAtY, uint32_t *border_hack
...
 {
-	int32_t iContact = 0;
-	bool fAnyContact = false;
-	int iContacts = 0;
-	BYTE fTurned = 0;
-	BYTE fRedirectYR = 0;
-	BYTE fNoAttach = 0;
+	bool has_moved = false;
+	bool has_contact = false;
+	int contact_count = 0;
```
isilkor: iirc this isn't a count but a bitmask.

src/object/C4Object.h
```diff
@@ -247,7 +248,7 @@ class C4Object: public C4PropListNumbered
 	void GetOCFForPos(int32_t ctx, int32_t cty, DWORD &ocf) const;
 	bool CloseMenu(bool fForce);
 	bool ActivateMenu(int32_t iMenu, int32_t iMenuSelect=0, int32_t iMenuData=0, int32_t iMenuPosition=0, C4Object *pTarget=nullptr);
-	int32_t ContactCheck(int32_t atx, int32_t aty, uint32_t *border_hack_contacts=nullptr, bool collide_halfvehic=false);
+	int32_t ContactCheck(int32_t at_x, int32_t at_y, uint32_t *border_hack_contacts=nullptr, bool collide_halfvehic = false);
```
isilkor: If you add whitespace to `collide_halfvehic`, you should do it to `border_hack_contacts` too

isilkor:
In re b69dfec: (Most of) those comments are fine starting with a lower-case letter because they don't start a complete sentence.

src/object/C4Movement.cpp
gitMarky: Indeed. Rename to "contact_info" maybe? "contact_bitmask" seems too strange. Also, it seems that t_contact should be "temp_contact" or something like that.

gitMarky:
Also, what about moving the MovementDigFreeTargetArea() from C4Object::DoMovement() to C4Object::ExecMovement()?
It feels strange to call this in a function that should do movement, and it does a best guess where the object might end up, based on velocity, and digs free there.

src/object/C4Movement.cpp
```diff
@@ -184,122 +182,99 @@ int32_t C4Object::ContactCheck(int32_t iAtX, int32_t iAtY, uint32_t *border_hack
...
+	BYTE has_turned = 0;
+	BYTE redirected_force_from_ydir_to_rdir = 0;
```
gitMarky: The variable name is really long, if you have a better suggestion, please tell me so.

SvenTwo: I think it's better to err on the side of being more descriptive. The name looks fine to me.

src/object/C4Movement.cpp
```diff
@@ -184,122 +182,99 @@ int32_t C4Object::ContactCheck(int32_t iAtX, int32_t iAtY, uint32_t *border_hack
...
-	int32_t iContact = 0;
-	bool fAnyContact = false;
-	int iContacts = 0;
-	BYTE fTurned = 0;
-	BYTE fRedirectYR = 0;
-	BYTE fNoAttach = 0;
+	bool has_moved = false;
+	bool has_contact = false;
+	int contact_count = 0;
+	BYTE has_turned = 0;
```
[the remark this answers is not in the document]
SvenTwo: Most legacy code uses integers for booleans because the old serialization code would only understand integers.

Check for all uses before turning it into bool though to make sure that the integer value isn't actually used.

isilkor:
gitMarky wrote on 07.03.2019 18:41:
> Also, what about moving the MovementDigFreeTargetArea() from 
> C4Object::DoMovement() to C4Object::ExecMovement()?
> It feels strange to call this in a function that should do movement, and 
> it does a best guess where the object might end up, based on velocity, 
> and digs free there.
I think any functional changes should be part of a separate PR.

gitMarky:
> gitMarky wrote on 07.03.2019 18:41:
> Also, what about moving the MovementDigFreeTargetArea() from C4Object::DoMovement() to C4Object::ExecMovement()? It feels strange to call this in a function that should do movement, and it does a best guess where the object might end up, based on velocity, and digs free there.
> I think any functional changes should be part of a separate PR.

That is OK, I wanted an opinion before I change and try it out, though.

src/object/C4Movement.cpp
gitMarky: Well, any suggestion?

gitMarky: Renamed it now, take a look: https://github.com/openclonk/openclonk/pull/88/commits/64fec151a0375ffee77ac50e5fbedf0f3f3a0ebc

PR #89, opened 2019-03-15
Allow Material property for map script algorithms.

fohly:
Adds a new internal algorithm _C4MapScriptAlgoSetMaterial_ and uses it, when the _Material_ property is used for an algorithm.


isilkor:
Since Sven isn't available at the moment, I'm going to merge this. Thanks for the contribution!

PR #91, opened 2019-03-30
Refactor shape

gitMarky:
Preparation for adding vertex lines.

### DIFF
--- a/src/object/C4Shape.cpp
+++ b/src/object/C4Shape.cpp
@@ -343,10 +343,7 @@ out:
 			ciy = oldy;
 			if (!PathFreeIgnoreVehicle(cix, ciy, tx, ty) || !PathFreeIgnoreVehicle(cix, ciy, VtxX[cvtx + ld], VtxY[cvtx + ld]))
 			{
-				if (!PathFreeIgnoreVehicle(cix, ciy, tx, ty) || !PathFreeIgnoreVehicle(cix, ciy, VtxX[cvtx + ld], VtxY[cvtx + ld]))
-				{
-					return false; // Found no bend vertex
-				}
+				return false; // Found no bend vertex
 			}
 		}
 		// Insert bend vertex
--- a/src/object/C4Shape.cpp
+++ b/src/object/C4Shape.cpp
@@ -530,16 +530,16 @@ int32_t C4Shape::GetVertexY(int32_t iVertex)
 	return VtxY[iVertex];
 }
 
-void C4Shape::CopyFrom(C4Shape shape_from, bool copy_vertices, bool copy_vertices_from_self)
+void C4Shape::CopyFrom(C4Shape source, bool copy_vertices, bool copy_vertices_from_self)
 {
 	if (copy_vertices)
 	{
 		// Truncate / copy vertex count
-		VtxNum = (copy_vertices_from_self ? std::min<int32_t>(VtxNum, C4D_VertexCpyPos) : shape_from.VtxNum);
+		VtxNum = (copy_vertices_from_self ? std::min<int32_t>(VtxNum, C4D_VertexCpyPos) : source.VtxNum);
 
 		// Restore vertices from back of own buffer (retaining count)
 		int32_t iCopyPos = (copy_vertices_from_self ? C4D_VertexCpyPos : 0);
-		C4Shape &vertices_from = (copy_vertices_from_self ? *this : shape_from);
+		C4Shape &vertices_from = (copy_vertices_from_self ? *this : source);
 
 		memcpy(VtxX,           vertices_from.VtxX + iCopyPos,           VtxNum * sizeof(*VtxX));
 		memcpy(VtxY,           vertices_from.VtxY + iCopyPos,           VtxNum * sizeof(*VtxY));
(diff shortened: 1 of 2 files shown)


src/object/C4Shape.cpp
```diff
@@ -25,180 +25,229 @@
 #include "landscape/C4Landscape.h"
 #include "landscape/C4Material.h"
 
+
```
isilkor: We don't use double blank lines anywhere else.

src/object/C4Shape.cpp
```diff
@@ -25,180 +25,229 @@
...
+	{
+		for (int32_t i = 0; i < VtxNum; i++)
+		{
+			VtxY[i] = ScaledByCon(VtxY[i], iCon);
+		}
+	}
 }
 
+
+inline bool C4Shape::ScaledByCon(int32_t value, int32_t con)
```
isilkor: This function doesn't require any members of C4Shape and can be made static.

isilkor: Actually, since it's not used outside of C4Shape.cpp at all, you can even make it a function that's local to C4Shape.cpp.

src/object/C4Shape.cpp
```diff
@@ -273,250 +331,346 @@ bool C4Shape::LineConnect(int32_t tx, int32_t ty, int32_t cvtx, int32_t ld, int3
...
 		if (!found)
 		{
-			// try bending directly at path the line took
-			// allow going through vehicle in this case to allow lines through castles and elevator shafts
+			// Try bending directly at path the line took
+			// Allow going through vehicle in this case to allow lines through castles and elevator shafts
 			cix = oldx;
 			ciy = oldy;
-			if (!PathFreeIgnoreVehicle(cix,ciy,tx,ty) || !PathFreeIgnoreVehicle(cix,ciy,VtxX[cvtx+ld],VtxY[cvtx+ld]))
-				if (!PathFreeIgnoreVehicle(cix,ciy,tx,ty) || !PathFreeIgnoreVehicle(cix,ciy,VtxX[cvtx+ld],VtxY[cvtx+ld]))
```
isilkor: What's even happening here? Isn't this just the same test done twice?

src/object/C4Shape.cpp
```diff
@@ -273,250 +331,346 @@ bool C4Shape::LineConnect(int32_t tx, int32_t ty, int32_t cvtx, int32_t ld, int3
...
-	if (!Inside<int32_t>(iVertex,0,VtxNum-1)) return 0;
+	if (!Inside<int32_t>(iVertex, 0, VtxNum - 1))
+	{
+		return 0;
+	}
 	return VtxY[iVertex];
 }
 
-void C4Shape::CopyFrom(C4Shape rFrom, bool bCpyVertices, bool fCopyVerticesFromSelf)
+void C4Shape::CopyFrom(C4Shape shape_from, bool copy_vertices, bool copy_vertices_from_self)
```
isilkor: Just `from` (or `source`) is fine, we know it's a shape from its type.

src/object/C4Shape.cpp
gitMarky: How do I do that? Static is obvious, but I always get a compiler warning that the function is not used then (I am not C++ native ;))

isilkor: Remove it from the header and replace the declaration with
```cpp
static int32_t ScaledByCon(int32_t value, int32_t con)
```

src/object/C4Shape.cpp
gitMarky: At first I thought that it wants to get certain coordinates, but those are the 5th and 6th parameters and they are unused here. 
It seems that this actually does nothing then. I can remove the double check and see if it breaks something.

src/object/C4Shape.cpp
gitMarky: Done

src/object/C4Shape.cpp
gitMarky: Removed them again

src/object/C4Shape.cpp
gitMarky: Yes, wow, I do not even know how I could miss that. Although, I wanted to make it inline, but the compiler cannot resolve the static inline function - OK, that was resolved by placing the function before the first use (had nothing to do with inline, just with the way C++ works)

gitMarky: Also, I noticed that the collection zone (and likely the entrace, etc.) do not scale with Con. For structures that makes sense, but I think that for objects with StretchGrowth these should change with Con, too. Might be worth an issue in our bug tracker. Apart from that, the changes are done.

gitMarky:
Do you mind the reviews? Actually, the deeper I dig into the object code the more I find things that I want to "fix" immediately (well, just code style changes mostly, as was the case with the previous "preparation" pull requests). Should I file individual PRs, or can I just do a huge one? (Maybe in two steps: only style changes as one PR, variable scopes and other minor changes that may have side effects, but do not actually change the intended functionality as a second PR)?

I realize, that reviewing everything takes some time and the actual effort of combing through everything may not be worth it - however, at the moment we have few people working on the engine, so it is unlikely that we get a lot of conflicts, and I like cleaning up things.

isilkor:
I don't mind the reviews, but keep in mind that you're currently working 
on code I'm not exactly familiar with, so for changes more intrusive 
than plain refactoring I can't really contribute much.

gitMarky:
That is OK, because if anything I'd do a lot of refactoring first. As I mentioned in one of the comments: establishing a naming convention that is more or less consistent throughout the engine code will be helpful, and only then real changes make sense (well, simple "I'll hack my stuff into it" is also an option, but why read the same weird names "ctx, cty" and the like repeatedly?

isilkor:
> That is OK, because if anything I'd do a lot of refactoring first. As I 
> mentioned in one of the comments: establishing a naming convention that 
> is more or less consistent throughout the engine code will be helpful, 

FWIW, I think that "obj" is fine to denote the one object a function is 
operating on (just like we're using "i" for loop indices instead of 
"index"). When some code handles more than one object, they should get 
better names so people don't have to try and figure out which one is 
"obj" and which one is "obj2" (or similar).

fohly:
This breaks object rotation.

gitMarky:
Is it a general problem or in specific situations?

fohly:
always, as far as i can tell

PR #93, opened 2019-04-01
Refactor generic

gitMarky:
Lots of code style changes, some simplifications. Wanted to do C4Object.cpp first, but that is not really viable - I prefer smaller files for that, also found lots of things that I wanted to change in that file, but splitting it up, similar to C4Movement.cpp should be done first.

### DIFF
--- a/src/object/C4Id.h
+++ b/src/object/C4Id.h
@@ -34,7 +34,7 @@ private:
 	static NamesList names;
 	void assign(const std::string &s);
 	template<size_t N>
-	explicit C4ID(const char (&s)[N]) { assign(s); } // TODO: v is not assigned
+	explicit C4ID(const char (&s)[N]) { assign(s); } // @suppress("Class members should be properly initialized"): The call to assign() initializes 'v'
 public:
 	static const C4ID None; // Invalid ID
 	DEPRECATED static const C4ID Flag;
--- a/src/object/C4Id.h
+++ b/src/object/C4Id.h
@@ -34,13 +34,13 @@ private:
 	static NamesList names;
 	void assign(const std::string &s);
 	template<size_t N>
-	explicit C4ID(const char (&s)[N]) { assign(s); }
+	explicit C4ID(const char (&s)[N]) { assign(s); } // TODO: v is not assigned
 public:
 	static const C4ID None; // Invalid ID
-		DEPRECATED static const C4ID Flag;
+	DEPRECATED static const C4ID Flag;
 	DEPRECATED static const C4ID Conkit; // Construction kit
 	static const C4ID Clonk;
-	static const C4ID Melee;
+	DEPRECATED static const C4ID Melee;
 	static const C4ID Bubble;
 	static const C4ID EditorBase;
 
@@ -76,8 +76,8 @@ public:
 	// Compare names instead of v directly so that a sequence of IDs is synchronous
 	inline bool operator ==(const C4ID &other) const { return names[v] == names[other.v]; }
 	inline bool operator !=(const C4ID &other) const { return names[v] != names[other.v]; }
-	inline bool operator <(const C4ID &other) const { return names[v] < names[other.v]; }
-	inline bool operator >(const C4ID &other) const { return names[v] > names[other.v]; }
+	inline bool operator <(const C4ID &other)  const { return names[v] < names[other.v];  }
+	inline bool operator >(const C4ID &other)  const { return names[v] > names[other.v];  }
 	inline bool operator <=(const C4ID &other) const { return names[v] <= names[other.v]; }
 	inline bool operator >=(const C4ID &other) const { return names[v] >= names[other.v]; }
 
--- a/src/object/C4ObjectInfoList.cpp
+++ b/src/object/C4ObjectInfoList.cpp
@@ -113,7 +113,6 @@ void C4ObjectInfoList::MakeValidName(char *sName)
 	size_t namelen = SLen(sName);
 	for (int32_t identifier = 2; NameExists(sName); identifier++)
 	{
-		assert(identifier > 1);
 		sprintf(number_suffix, " %d", identifier);
 		SCopy(number_suffix, sName + std::min(namelen, C4MaxName - SLen(number_suffix)));
 	}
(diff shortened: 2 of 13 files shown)


src/object/C4Id.h
```diff
@@ -34,13 +34,13 @@ class C4ID
 	static NamesList names;
 	void assign(const std::string &s);
 	template<size_t N>
-	explicit C4ID(const char (&s)[N]) { assign(s); }
+	explicit C4ID(const char (&s)[N]) { assign(s); } // TODO: v is not assigned
```
isilkor: It is assigned, either in https://github.com/openclonk/openclonk/blob/53fccd01d1ee9f77c14ac6c96b6065b01a0b9da9/src/object/C4Id.cpp#L45 or in https://github.com/openclonk/openclonk/blob/53fccd01d1ee9f77c14ac6c96b6065b01a0b9da9/src/object/C4Id.cpp#L49

src/object/C4ObjectScript.cpp
```diff
@@ -86,53 +92,72 @@ static void FnDeathAnnounce(C4Object *Obj)
...
+		sprintf(idDeathMsg, "IDS_OBJ_DEATH%d", 1 + UnsyncedRandom(MaxDeathMsg));
 		GameMsgObject(FormatString(LoadResStr(idDeathMsg), Obj->GetName()).getData(), Obj);
 	}
 }
 
 static bool FnGrabContents(C4Object *Obj, C4Object *from)
 {
-	if (!from) return false;
-	if (Obj == from) return false;
+	if (!from && Obj == from)
```
isilkor: This should be an `||`.

src/object/C4Id.h
gitMarky: Okay, maybe the comment was misleading. The C++ code analyzer in Eclipse complains that "Member 'v' was not initialized in this constructor"

src/object/C4ObjectScript.cpp
```diff
@@ -1703,8 +2119,10 @@ static long FnGetObjectStatus(C4Object *Obj)
 static bool FnAdjustWalkRotation(C4Object *Obj, long iRangeX, long iRangeY, long iSpeed)
 {
 	// must be rotateable and attached to solid ground
-	if (!Obj->Def->Rotateable || ~Obj->Action.t_attach&CNAT_Bottom || Obj->Shape.AttachMat == MNone)
+	if (!Obj->Def->Rotateable || ~(Obj->Action.t_attach & CNAT_Bottom) || Obj->Shape.AttachMat == MNone)
```
isilkor: Those parentheses are incorrect; the negation binds to `Obj->Action.t_attach`, not to the entire masking result.

src/object/C4Id.h
```diff
@@ -34,13 +34,13 @@ class C4ID
...
 public:
 	static const C4ID None; // Invalid ID
-		DEPRECATED static const C4ID Flag;
+	DEPRECATED static const C4ID Flag;
```
lluchs: Are these still used anywhere? We should just remove them otherwise...

gitMarky: Flag and Conkit are not used, Melee is used by C4SGame::IsMelee(), but imo it can be removed there, because we no longer have such an ID. The other solution is updating the ID to the current melee goal.

src/object/C4ObjectInfoList.cpp
```diff
@@ -39,95 +39,122 @@ C4ObjectInfoList::~C4ObjectInfoList()
...
+		return false;
+	}
+	pInfo->Next = First;
+	First = pInfo;
 	return true;
 }
 
 void C4ObjectInfoList::MakeValidName(char *sName)
 {
 	char tstr[_MAX_PATH];
```
lluchs: `_MAX_PATH` doesn't make sense at all here.

src/object/C4ObjectList.cpp
```diff
@@ -163,8 +213,12 @@ bool C4ObjectList::Add(C4Object *nObj, SortType eSort, C4ObjectList *pLstSorted)
 			// Sort by master list?
 			if (pLstSorted)
 			{
-				cPrev = nullptr; cLnk = First;
-				while(cLnk && (!cLnk->Obj->Status || cLnk->Obj->Unsorted)) cLnk = cLnk->Next;
+				cPrev = nullptr;
+				cLnk = First;
+				while(cLnk && (!cLnk->Obj->Status || cLnk->Obj->Unsorted))
```
lluchs: missing space `while `

src/object/C4ObjectList.cpp
```diff
@@ -284,10 +366,17 @@ bool C4ObjectList::Remove(C4Object *pObj)
...
-	Mass-=pObj->Mass; if (Mass<0) Mass=0;
+	Mass -= pObj->Mass;
+	if (Mass < 0)
+	{
+		Mass = 0;
+	}
 
 #if defined(_DEBUG)
-	if (GetLink(pObj)) BREAKPOINT_HERE;
+	if (GetLink(pObj))
```
lluchs: I think this should just be `assert(!GetLink(pObj))`?

src/object/C4ObjectList.cpp
```diff
@@ -574,42 +767,74 @@ void C4ObjectList::Sort()
...
 	}
 	while (!fSorted);
 }
 
 void C4ObjectList::RemoveLink(C4ObjectLink *pLnk)
 {
-	if (pLnk->Prev) pLnk->Prev->Next=pLnk->Next; else First=pLnk->Next;
-	if (pLnk->Next) pLnk->Next->Prev=pLnk->Prev; else Last=pLnk->Prev;
+	if (pLnk->Prev)
+	{
```
lluchs: I actually like the short variant better here because it shows the symmetry, but whatever.

In any case, it might be worth it to standardize the normal "link" variable name in this file. `cLnk` seems to be the most common one (whatever the c stands for), but there's also `clnk` (which is very confusing because it would stand for "Clonk" everywhere else) and `pLnk` (here).

gitMarky: I'll change it back to the old notation then.

src/object/C4ObjectInfoList.cpp
gitMarky: what should we do with it?

gitMarky: To be honest, I do not understand how the function makes a valid name.

src/object/C4Id.h
gitMarky: Whatever the solution, I'd merge the PR first and then remove the deprecated IDs on master.

PR #94, opened 2019-04-02
Split C4Object.cpp into more digestable parts

gitMarky:
The huge file was confusing and makes it hard to work on a specific part without also causing merges for anyone who works on something completely different.
In this PR I tried to split up the file into thematic parts, without changing code style or functionality.
"git blame" will blame me for changes in the new files, but calling it with "-M -C" shows the original commit.


gitMarky:
Luchs had a look at it yesterday.

PR #95, opened 2019-04-05
Fix bug #1251

gitMarky:
Fix for https://bugs.openclonk.org/view.php?id=1251

Will have a conflict with PR#94, I suppose, because there is no formatting change here yet.


src/object/C4DefList.cpp
```diff
@@ -140,7 +140,7 @@ int32_t C4DefList::Load(C4Group &hGroup, DWORD dwLoadWhat,
 		}
 
 	// load additional system scripts for def groups only
```
lluchs: Comment is wrong now.

PR #96, opened 2019-04-10
Removed bridge action

gitMarky:
As requested by myself here https://bugs.openclonk.org/view.php?id=2058


gitMarky:
I kept the docu, because it may confuse people who knew that it was there? But yes, there is not much use for it

PR #108, opened 2019-05-02
Fix same object being removed multiple times (#2048)

gitMarky:
See https://bugs.openclonk.org/view.php?id=2048

### DIFF
--- a/src/object/C4Object.cpp
+++ b/src/object/C4Object.cpp
@@ -227,6 +227,8 @@ void C4Object::AssignRemoval(bool exit_contents)
 	}
 	Status = C4OS_DELETING;
 
+	// ----------------------------------------------------------
+	// Debugging
 	if (Config.General.DebugRec)
 	{
 		C4RCCreateObj rc;
@@ -255,7 +257,6 @@ void C4Object::AssignRemoval(bool exit_contents)
 		}
 	}
 
-
 	// ----------------------------------------------------------
 	// Destruction call
 	Call(PSF_Destruction);
@@ -264,7 +265,6 @@ void C4Object::AssignRemoval(bool exit_contents)
 	{
 		return;
 	}
-
 
 	// ----------------------------------------------------------
 	// Remove all effects (extinguishes as well)
@@ -280,12 +280,19 @@ void C4Object::AssignRemoval(bool exit_contents)
 
 	// ----------------------------------------------------------
 	// Remove particles, no destruction check necessary
-	ClearParticleLists();
+	ClearParticleLists(); // TODO: Move this to FinishRemoval(), too?
 
 	// ----------------------------------------------------------
 	// Action idle
 	SetAction(nullptr);
-	// Action callbacks might have deleted the object already
+
+	// ----------------------------------------------------------
+	// Actual cleanup
+	FinishRemoval(exit_contents);
+}
+
+void C4Object::FinishRemoval(bool exit_contents)
+{
 	if (Status == C4OS_DELETED)
 	{
 		return;
@@ -299,15 +306,7 @@ void C4Object::AssignRemoval(bool exit_contents)
 		Status = C4OS_NORMAL;
 		::Objects.Add(this);
 	}
-	FinishRemoval(exit_contents);
-}
 
-void C4Object::FinishRemoval(bool exit_contents)
-{
-	if (Status == C4OS_DELETED)
-	{
-		return;
-	}
 	Status = C4OS_DELETED;
 	// count decrease
 	Def->Count--;
--- a/src/object/C4Object.cpp
+++ b/src/object/C4Object.cpp
@@ -225,9 +225,12 @@ void C4Object::AssignRemoval(bool exit_contents)
 	{
 		return;
 	}
+	// Set status to deleting, so that callbacks in this function that might delete
+	// the object do not delete it a second time.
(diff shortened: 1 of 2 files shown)


src/object/C4Object.h
```diff
@@ -101,6 +101,7 @@ class C4Object: public C4PropListNumbered
 	void Splash();
 	void RemoveSolidMask(bool fBackupAttachment); // Remove solid mask data, if existing
 	void MovementDigFreeTargetArea(); // Dig the area free, according to action data
+	inline void FinishRemoval(bool exit_contents);
```
walachey: Why inline? In this case it's not needed to be able to include it (as it's defined in the .cpp file). And compilers will decide themselves anyway what to actually inline and what not.
See https://stackoverflow.com/questions/29796264/is-there-still-a-use-for-inline

src/object/C4Object.cpp
```diff
@@ -231,37 +231,70 @@ void C4Object::AssignRemoval(bool exit_contents)
...
+		{
+			strncpy(rc.id, Def->GetName(), 32+1);
+		}
+		rc.x = GetX();
+		rc.y = GetY();
+		rc.ownr = Owner;
 		AddDbgRec(RCT_DsObj, &rc, sizeof(rc));
 	}
+
+	// ----------------------------------------------------------
```
walachey: Do we do this weird formatting with comments in other parts of the codebase, too?

src/object/C4Object.cpp
```diff
@@ -280,12 +280,19 @@ void C4Object::AssignRemoval(bool exit_contents)
 
 	// ----------------------------------------------------------
 	// Remove particles, no destruction check necessary
-	ClearParticleLists();
+	ClearParticleLists(); // TODO: Move this to FinishRemoval(), too?
```
walachey: That kind of depends on what the functions are supposed to do, which isn't clear from the names (AssignRemoval vs. FinishRemoval) - and they are also not commented. So for an outsider it is very unclear how to answer this question

src/object/C4Object.cpp
gitMarky: At least in some parts. Movement code had this to begin with, and I removed some of it. Can remove it again, but for a short time it helped me to better see the structure here.

Anyway, the more relevant question here is: Do we still need the if (status ...) return; statements? I think not, because the callbacks cannot remove the object anymore!

src/object/C4Object.cpp
gitMarky: Yes, the TODO was more or less for the reviewer. More on that in the other comment, where FinishRemoval() was introduced.

src/object/C4Object.h
gitMarky: Thanks for the article! So, the purpose was to separate the parts "Start removing the object and do all the callbacks" (=AssignRemoval) and "Actually remove the object" (=FinishRemoval). Making it inline had the intention of telling the compiler to not make it an actual separate function.

isilkor: > Why inline? In this case it's not needed to be able to include it (as it's defined in the .cpp file).
It wouldn't be required even if it was defined here because method definitions inside the class definition always have inline linkage.

src/object/C4Object.cpp
```diff
@@ -225,9 +225,12 @@ void C4Object::AssignRemoval(bool exit_contents)
 	{
 		return;
 	}
+	// Set status to deleting, so that callbacks in this function that might delete
+	// the object do not delete it a second time.
+	// TODO: This causes problems if the status should change during one of those
+	// callbacks, or if the status was C4OS_INACTIVE
```
gitMarky: See the additional TODO:
Either a) we keep the old "if (Status == C4OS_DELETED) return" calls to be safe and remove the C4OS_DELETING again; or
b) we make "object is being deleted" a bool that is independent from the status?

I guess the following options are not a valid solution:
c) Make the status bitwise, instead of exclusive. So you could have Status = C4OS_DELETING | C4OS_INACTIVE. This is not really safe, because I assume that status was assigned directly in other places, meaning that we have to adjust other files, too
d) Set the status C4OS_DELETED at the beginning of the function and remove the C4OS_DELETING again. This was probably not done, because the rest of the engine assumes that the object is gone when the status is C4OS_DELETED, so the callbacks might not do their stuff properly if the status is set beforehand?

src/object/C4Object.cpp
walachey: Even the reviewer needs to know what the function is supposed to do - and that should be reflected in the source. Either with a descriptive function name or with a comment. Because if it's ambiguous where something should go, you can be sure that someone puts it into the wrong place (e.g. the next time someone needs a place for a cleanup method and thinks about whether to put it into FinishRemoval or AssignRemoval).
Maybe if the functions have a different premise (e.g. a certain object status) that can be violated if people use it incorrectly, this premise could be reflected in some assertions?

gitMarky: Yes, the comments were added to the header file in the latest commit. From my point of view, since clearing the particles does not rely on callbacks and does not change the object state, it makes more sense to do it later when the object is actually removed. 

The more pressing issue here is, that the PR implementation breaks scenario sections, see the comment after my additional commits. I am unsure how to address this.

At first, the fix seemed pretty simple, but it is not :)

PR #109, opened 2019-05-03
Fixed the editor preview for Decoration objects.

fohly:
Could moving MeshTransformation from Construction to Definition break any scenarios?

Also, TreeTrunks and WineBarrel still have an offset preview, due to uncommon DefCore. (Offset is not half width or height, resp.)
I believe that is a bug in the editor preview code.
Fixing their DefCore would definitely break placement in existing scenarios.


planet/Decoration.ocd/Village.ocd/Bag.ocd/Script.c
```diff
@@ -2,6 +2,10 @@
 	@author Dustin Neß (dness.de)
 */
 
-protected func Construction() {
-SetProperty("MeshTransformation", Trans_Mul(Trans_Rotate(RandomX(-180,180),0,10), Trans_Scale(RandomX(130,200))));
-}
\ No newline at end of file
+public func Definition(proplist def) {
+	def.MeshTransformation = Trans_Scale(165); // average scale
```
gitMarky: How is 165 the average of RandomX(787, 1212)? (You changed the random values in the Construction() call)

planet/Decoration.ocd/Village.ocd/BridgeRustic.ocd/Script.c
```diff
@@ -19,3 +19,5 @@ protected func Construction()
 
 	return true;
 }
+
+local Visibility = VIS_Editor;
```
gitMarky: That would make it visible in the editor, but invisible in the game. Sounds unintended.

planet/Decoration.ocd/Village.ocd/Crate.ocd/Script.c
```diff
@@ -2,8 +2,14 @@
...
\ No newline at end of file
+	def.MeshTransformation = Trans_Scale(10, 10, 100);
+}
+
+public func Construction() {
+	// Note: We apply the scaling after the rotation, which is the "wrong"
+	//       order. However, due to what I believe is an engine bug,
+	//       scaling down too much along the z axis makes the object disappear.
+	SetProperty("MeshTransformation", Trans_Mul(GetID().MeshTransformation, Trans_Rotate(RandomX(-180,180), 0, 1, 0)));
+}
```
gitMarky: I think that the definition should just define the scale here, as in the changes to e.g. the candle. Why give all editor previews a fixed rotation if you change that to another random rotation once the object is placed in the game?

planet/Decoration.ocd/Village.ocd/FenceRustic.ocd/Script.c
```diff
@@ -2,7 +2,11 @@
...
 {
-	SetProperty("MeshTransformation", Trans_Mul(Trans_Rotate(RandomX(-35,35),0,10), Trans_Scale(60)));
-}
\ No newline at end of file
+	def.MeshTransformation = Trans_Scale(60);
+}
+
+public func Construction() {
+	SetProperty("MeshTransformation", Trans_Mul(Trans_Rotate(RandomX(-35,35),0,10), GetID().MeshTransformation));
+}
```
gitMarky: By the way, the "protected func" engine callbacks can be reduced to simply "func". "public func" is necessary only if you want this function to be displayed in the editor.

planet/Decoration.ocd/Village.ocd/Lantern.ocd/Script.c
```diff
@@ -12,9 +12,13 @@ protected func Initialize()
 	return SetAction("Shine");
 }
 
+public func Definition(proplist def) {
```
gitMarky: ``Could you please write the curly braces in the next line, as it is the case for all the other functions?

planet/Decoration.ocd/Village.ocd/TunnelSupport.ocd/Script.c
```diff
@@ -7,9 +7,13 @@
...
 
+public func Definition(proplist def) {
+	def.MeshTransformation = Trans_Rotate(90, 0, 1, 0);
+}
+
 func Construction()
 {
-	SetProperty("MeshTransformation", Trans_Rotate(RandomX(75, 105), 0,1,0));
+	SetProperty("MeshTransformation", Trans_Mul(Trans_Rotate(RandomX(-15, 15), 0,1,0), GetID().MeshTransformation));
 }
```
gitMarky: Here the separate rotation in the definition makes sense, keep it that way!

src/editor/C4EditCursor.cpp
```diff
@@ -880,7 +880,9 @@ void C4EditCursor::Draw(C4TargetFacet &cgo)
 	if (Mode == C4CNS_ModeCreateObject && has_mouse_hover && creator_def)
 	{
 		C4TargetFacet cgo_creator;
-		cgo_creator.Set(cgo.Surface, X + cgo.X - cgo.TargetX, Y + cgo.Y - cgo.TargetY,
+		cgo_creator.Set(cgo.Surface,
+			X + cgo.X - cgo.TargetX + creator_def->Shape.x + creator_def->Shape.Wdt / 2,
+			Y + cgo.Y - cgo.TargetY + creator_def->Shape.y + creator_def->Shape.Hgt / 2,
 			creator_def->Shape.Wdt, creator_def->Shape.Hgt, 0, 0, cgo.Zoom, 0, 0);
```
gitMarky: Could you explain what this does?

Also, just guessing now, I think that it works for objects where the offset is symmetrical i.e. Offset = -20, -5, Width = 40, Height = 10. What about an object with Offset = -10, -2, Width = 40, Height = 10?

planet/Decoration.ocd/Village.ocd/Bag.ocd/Script.c
fohly: 165 is the average of the old random range 130 to 200.
.165 * .787 ≈ .13
.165 * 1.212 ≈ .2

planet/Decoration.ocd/Village.ocd/BridgeRustic.ocd/Script.c
fohly: The object is transparent anyway.
I figured this would perhaps allow for optimization.

planet/Decoration.ocd/Village.ocd/Crate.ocd/Script.c
fohly: > I think that the definition should just define the scale here, as in the changes to e.g. the candle.

It is already the same way?

> Why give all editor previews a fixed rotation if you change that to another random rotation once the object is placed in the game?

The original code has the random rotation. And as far as I can tell, it is not possible to have a random rotation in the preview.

src/editor/C4EditCursor.cpp
fohly: This is to fix object preview where Offset is not (`-Width / 2`, `-Height / 2`)

I think the current code assumed that Offset is always `-Wdt / 2` (or Height, resp.)
So I basically add the real offset (which is negative usually) and compensate by adding `Wdt / 2`.
As far as I can tell, this looks right for all existing objects.

planet/Decoration.ocd/Village.ocd/Lantern.ocd/Script.c
fohly: Will do.

planet/Decoration.ocd/Village.ocd/Bag.ocd/Script.c
gitMarky: Ah, well, blame the github display then :D. The line is displayed as

SetProperty("MeshTransformation", Trans_Mul(Trans_Rotate(RandomX(-180,180),0,10), Trans_Scale(RandomX(787,1212))

and I need to scroll right in order to see the , GetID().MeshTransformation));

planet/Decoration.ocd/Village.ocd/BridgeRustic.ocd/Script.c
gitMarky: Transparent of invisible?

planet/Decoration.ocd/Village.ocd/Crate.ocd/Script.c
gitMarky: Ah, yes this one has scale only in the Definition() call. Why Trans_Scale(10, 10, 100) ? Should be Trans_Scale(10) only, imo.

src/editor/C4EditCursor.cpp
gitMarky: Add an explaining comment then, please

planet/Decoration.ocd/Village.ocd/BridgeRustic.ocd/Script.c
fohly: Yes?

planet/Decoration.ocd/Village.ocd/Crate.ocd/Script.c
fohly: It's explained in the comment above. Scaling down too much along the z axis makes the object disappear.

gitMarky: Then why was it 10 before? The Z axis should be the one that points towards the screen, so having this at 100 and x at 10 should create rectangular boxes in theory...

planet/Decoration.ocd/Village.ocd/BridgeRustic.ocd/Script.c
gitMarky: If it is invisible, why have it visible in the editor? If it is just has transparent but is otherwise visible, then it should not have VIS_Editor

planet/Decoration.ocd/Village.ocd/Crate.ocd/Script.c
fohly: At least for me, the crates didn't show up at all before.

And the scaling is applied in screen space, so it still looks like a cube.

planet/Decoration.ocd/Village.ocd/BridgeRustic.ocd/Script.c
fohly: I've removed it.

planet/Decoration.ocd/Village.ocd/Crate.ocd/Script.c
gitMarky: Still weird, but ok then!

PR #110, opened 2019-05-05
Fix two allocate-deallocate mismatches. Found by ASan.

fohly:


fohly:
As far as I can tell, this doesn't cause any other problems?

PR #112, opened 2019-06-08
Change Push/PopActionSpeed to not work like a stack anymore

gitMarky:
I am unsure about this - it fixes my bug report: https://bugs.openclonk.org/view.php?id=2067

On the other hand, it is unclear why the previous implementation was designed as a stack and why the issue from my bug report was deemed OK. Now, the implementation behaves differently and may actually break things, but I tried to implement this as close to the original as possible.

### DIFF
--- a/planet/Objects.ocd/Clonk.ocd/Animations.ocd/Script.c
+++ b/planet/Objects.ocd/Clonk.ocd/Animations.ocd/Script.c
@@ -873,52 +873,55 @@ func FxFallTimer(object target, effect, int timer)
 	@par action The new speed value.
 	@par relative If true, the speed value will be interpreted
 	              as speed/1000 of the original speed.
+	@par identifier If an identifier is specified,
+	                then the fixed speed value is scaled by
+	                speed/1000.
+	                Otherwise, the fixed speed value is set.
 */
-func PushActionSpeed(string action, int speed, bool relative)
+func PushActionSpeed(string action, int speed, any identifier)
+{
+	SetActionSpeed(action, speed, identifier);
+}
+
+/* Resets the named action to the previous one */
+func PopActionSpeed(string action, any identifier)
+{
+	SetActionSpeed(action, nil, identifier);
+}
+
+func SetActionSpeed(string action, int speed, any identifier)
 {
 	// Create an actmap that can be manipulated, if necessary
 	if (ActMap == this.Prototype.ActMap)
 	{
 		ActMap = { Prototype = this.Prototype.ActMap };
 	}
-	// Update the speed value.
-	// The old value is saved in the prototype and will be restored by PopActionSpeed
-	// E.g. PushActionSpeed("foo", 30), PushActionSpeed("foo", 40),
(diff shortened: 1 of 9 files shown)


walachey:
I like that you can now optionally also specify a factor by which to scale the speed (as opposed to setting it to a fixed value), that reduces some code duplication.
And I also think that using an identifier instead of the stack is a good idea.

However, I see three issues:
The first is that `%v` is mainly meant for debug purposes. I am not sure what happens now when different people have the game set to a different language or when you modify the passed proplist between two calls.
E.g. I think that
```
this.Name = "bar";
SetActionSpeed("Walk", 1, this);
this.Name = "foo";
PopActionSpeed("Walk", this);```
Might not work. Or it might not work if the two calls are separates by saving the game and the user switches the language in between (even when you pass a definition instead of an object). I strongly recommend to get rid of `%v`.

On the same note, I would get rid of allowing all types for `identifier`, because stuff like this can lead to very subtle and hard-to-find errors. In all the code changes you use an ID. If you want to keep using that, then I would recommend to change the parameter type to only accept ID and add a `FatalError` if it is something else. I would also recommend to not allow passing `nil`.

The last issue is that currently your `int speed` behaves differently according to whether you pass an ìdentifier` or not. I strongly recommend against that, too. This is basically guaranteed to lead to subtle bugs, it is highly unexpected for the developer (and thus another source of wasted time) and it's not even clear from the signature of the function.
I would recommend to either use two speed arguments (`int absolute, int relative`) or at the very least a flag (`int speed_adjustment, bool relative`).

What do you think?

gitMarky:
> The first is that `%v` is mainly meant for debug purposes. I am not sure what happens now when different people have the game set to a different language or when you modify the passed proplist between two calls.
Suggestion: The parameter accepts a string, but C4ID is converted to a string using Format("%i") then - other types result in FatalError. Unfortunately, we cannot write, "string|id" (or can we?).

> I would also recommend to not allow passing `nil`.
Not a problem basically, but see below...

> The last issue is that currently your `int speed` behaves differently according to whether you pass an `identifier` or not. 
Yes, I do not like that either, the intention was to reduce work for external packs that rely on the old behaviour. More below :)

> I would recommend to either use two speed arguments (`int absolute, int relative`) or at the very least a flag (`int speed_adjustment, bool relative`).
Being able to set an absolute value at all is a problem, in my opinion: Is there a use case for that? One that I could think of: Make crew members walk a specific speed in a scripted sequence that relies on that speed (everyone is walking somewhere slowly, or the script executes a jump that needs a specific speed so that the object lands at the desired position).
If this is not a a use case, lets throw it out. Otherwise, it should override all the other relative values, and you have to make it a highly conscious decision, not the default for the function.

Armag3ddon:
I don't think this (setting absolute values) is a specific use case anywhere.

Armag3ddon:
Yeah, but that's mainly the case because someone did the first implementation (probably bow) and then the code was just copied.

gitMarky:
Good, then I'll simplify this a bit and adjust the other fixed values

gitMarky:
Could someone have a look at the latest changes again?

PR #113, opened 2019-06-14
Add search bar to the scenario selection list (only visible if there'…

Fulgen301:
…s no folder map, as there's nothing to search for there)

This is essentially the same as LegacyClonk commit [7c131866b4dcf29f1c550d33139ff85dbe7675bb](https://git.maxmitti.tk/lc/commit/?id=7c131866b4dcf29f1c550d33139ff85dbe7675bb) by myself.


src/gui/C4Gui.h
```diff
@@ -1284,10 +1284,10 @@ namespace C4GUI
 
 	private:
 		void Deselect(); // clear selection range
-		void ClearText();    // remove all the text
 
 	public:
 		bool InsertText(const char *szText, bool fUser); // insert text at cursor pos (returns whether all text could be inserted)
+		void ClearText();    // remove all the text
```
gitMarky: Not sure about that.. maybe "friend" the class that uses ClearText() and leave it private?

src/gui/C4StartupScenSelDlg.cpp
```diff
@@ -1442,6 +1442,21 @@ C4StartupScenSelDlg::C4StartupScenSelDlg(bool fNetwork) : C4StartupDlg(LoadResSt
 	pScenSelCaption = new C4GUI::Label("", caBookLeft.GetFromTop(rScenSelCaptionFont.GetLineHeight()), ACenter, ClrScenarioItem, &rScenSelCaptionFont, false);
 	pSheetBook->AddElement(pScenSelCaption);
 	pScenSelCaption->SetToolTip(LoadResStr("IDS_DLGTIP_SELECTSCENARIO"));
+
+	// search bar
+	const char *labelText = LoadResStr("IDS_DLG_SEARCH");
+	int32_t width = 100, height;
```
gitMarky: Height is either unused, or initialized by GetTextExtent? Anyway, please define variables in separate lines, and add a comment why they are not initialized, but used in later lines.

src/gui/C4StartupScenSelDlg.cpp
```diff
@@ -1442,6 +1442,21 @@ C4StartupScenSelDlg::C4StartupScenSelDlg(bool fNetwork) : C4StartupDlg(LoadResSt
...
+	::GraphicsResource.TextFont.GetTextExtent(labelText, width, height, true);
+	C4GUI::ComponentAligner caSearchBar(caBookLeft.GetFromBottom(height), 0, 0);
+	auto *searchLabel = new C4GUI::WoodenLabel(labelText, caSearchBar.GetFromLeft(width + 10), C4GUI_Caption2FontClr, &::GraphicsResource.TextFont);
```
gitMarky: Why did you use the auto keyword here?

src/gui/C4StartupScenSelDlg.cpp
```diff
@@ -1442,6 +1442,21 @@ C4StartupScenSelDlg::C4StartupScenSelDlg(bool fNetwork) : C4StartupDlg(LoadResSt
...
+	searchLabel->SetToolTip(LoadResStr("IDS_DLGTIP_SEARCHLIST"));
+	pSheetBook->AddElement(searchLabel);
+
+	searchBar = new C4GUI::CallbackEdit<C4StartupScenSelDlg>(caSearchBar.GetAll(), this, &C4StartupScenSelDlg::OnSearchBarEnter);
```
gitMarky: I'd prefer the variable declaration here, if this is the first time that you define searchBar.

Update: Well, github lists the .cpp file before the .h file, now it is clear to me that the searchBar is a field.

src/gui/C4StartupScenSelDlg.cpp
```diff
@@ -1499,9 +1514,7 @@ C4StartupScenSelDlg::C4StartupScenSelDlg(bool fNetwork) : C4StartupDlg(LoadResSt
 	SetFocus(pScenSelList, false);
 
 	// key bindings
-	C4CustomKey::CodeList keys;
-	keys.emplace_back(K_BACK); keys.emplace_back(K_LEFT);
-	pKeyBack = new C4KeyBinding(keys, "StartupScenSelFolderUp", KEYSCOPE_Gui,
+	pKeyBack = new C4KeyBinding(C4KeyCodeEx(K_LEFT), "StartupScenSelFolderUp", KEYSCOPE_Gui,
```
gitMarky: I do not have experience with the GUI stuff, could you explain what you did there? Apparently now the K_BACK is not used anymore - what effect does that have?
From my naive point of view the change does not look equivalent to the previous version.

src/gui/C4StartupScenSelDlg.cpp
```diff
@@ -1616,8 +1629,15 @@ void C4StartupScenSelDlg::UpdateList()
 		for (C4ScenarioListLoader::Entry *pEnt = pScenLoader->GetFirstEntry(); pEnt; pEnt = pEnt->GetNext())
 		{
 			if (pEnt->IsHidden()) continue; // no UI entry at all for hidden items
-			ScenListItem *pEntItem = new ScenListItem(pScenSelList, pEnt);
-			if (pEnt == pOldSelection) pScenSelList->SelectEntry(pEntItem, false);
+			if (!SLen(searchBar->GetText()) || SSearchNoCase(pEnt->GetName().getData(), searchBar->GetText()))
```
gitMarky: What does the if/else block do here? As far as I understand it: Search bar is clear, or found an item that matches the text => old behaviour. Otherwise => Clear "old selection".

src/gui/C4StartupScenSelDlg.h
```diff
@@ -474,6 +475,11 @@ class C4StartupScenSelDlg : public C4StartupDlg
 	void OnSelDblClick(class C4GUI::Element *pEl) { DoOK(); }
 	void OnButtonScenario(C4GUI::Control *pEl);
 
+	C4GUI::Edit::InputResult OnSearchBarEnter(C4GUI::Edit *edt, bool fPasting, bool fPastingMore)
+	{
+		UpdateList(); return C4GUI::Edit::IR_Abort;
```
gitMarky: Return in next line, please

src/gui/C4StartupScenSelDlg.cpp
isilkor: This way you don't need to repeat the type.

src/gui/C4Gui.h
isilkor: It's fine this way. I don't understand why ClearText was private in the first place.

lluchs: I think visibility in the C4Gui stuff is strictly by actual use and not to make a sane interface...

src/gui/C4StartupScenSelDlg.cpp
```diff
@@ -1616,8 +1629,15 @@ void C4StartupScenSelDlg::UpdateList()
...
+			{
+				ScenListItem *pEntItem = new ScenListItem(pScenSelList, pEnt);
```
walachey: Where is this deleted? I would highly recommend using smart pointers whenever possible in new code.
(Even if this might not be one of the location where it actually is possible)

src/gui/C4StartupScenSelDlg.cpp
Fulgen301: K_BACK is the backspace key which could be used to switch back the same way you could (and still can) do it with the left arrow key - however, it also switched back when the search bar had the focus, making text deletion impossible.

src/gui/C4StartupScenSelDlg.cpp
Fulgen301: Indeed.

src/gui/C4StartupScenSelDlg.h
Fulgen301: I copy-pasted that code from C4StartupNetDlg.

src/gui/C4StartupScenSelDlg.cpp
Fulgen301: [Here.](https://github.com/openclonk/openclonk/blob/master/src/gui/C4StartupScenSelDlg.cpp#L1588)
Yes, smart pointers are great, but I don't see the benefit in rewriting working code just to use them.

src/gui/C4StartupScenSelDlg.cpp
Fulgen301: >Anyway, please define variables in separate lines
Why? [The style guidelines don't forbid it.](https://wiki.openclonk.org/w/Style_Guidelines#Data_Types)

lluchs:
Code LGTM so far.

Some feedback for the feature itself:

- The tooltip translation doesn't work for me, I see "IDS_DLGTIP_SEARCHLIST" when hovering over the search input (language is English).
- There should be a message if there are no search results.
- Maybe it should also indicate that it's showing a filtered list. Something like "Search results:" on top?
- It's not searching (recursively) in scenario folders. This is not necessarily problematic for the first iteration of this feature, but it's a bit unintuitive. I think we should add this in the future.

src/gui/C4StartupScenSelDlg.cpp
gitMarky: I updated the style guidelines :D

The reason is easy, though: Although the code compiles, I do not know whether:
- You forgot assigning a value.
- You declared a variable but do not use it (especially applies if there are multiple in a line).
These are things that a code analyzer can easily tell you, but I do not have that when I am looking just at the code. Having the variable declarations in separate lines will point your attention towards them, especially unassigned ones.

In the updated guidelines I decided that multiple declarations of the same type within a line are OK, though, if no value is assigned. Any opinions on that?

src/gui/C4StartupScenSelDlg.cpp
gitMarky: Yes, sure, but is that such a problem?

src/gui/C4StartupScenSelDlg.cpp
gitMarky: Ok, makes sense

src/gui/C4StartupScenSelDlg.cpp
gitMarky: Could you add a comment that describes that then?

src/gui/C4StartupScenSelDlg.cpp
isilkor: Yes. http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-auto

src/gui/C4StartupScenSelDlg.cpp
isilkor: I'd rather have everything assigned a value. The compiler can eliminate dead stores. GetTextExtent should probably return a std::pair (or maybe a new Dimension type).

src/gui/C4StartupScenSelDlg.cpp
isilkor: > Where is this deleted? I would highly recommend using smart pointers whenever possible in new code.
I'm usually hugely in favor of using more smart pointers, but this isn't new code. Also whoever wants to use smart pointers in C4Gui probably needs to start at the top of the object hierarchy.

Fulgen301:
>The tooltip translation doesn't work for me, I see "IDS_DLGTIP_SEARCHLIST" when hovering over the search input (language is English).

Whoops, fixed.

>Maybe it should also indicate that it's showing a filtered list. Something like "Search results:" on top?
I find it a bit redundant, as the player can check whether the searchbar's empty. But then again, this might not work if he entered whitespace...

>There should be a message if there are no search results.
It's not searching (recursively) in scenario folders. This is not necessarily problematic for the first iteration of this feature, but it's a bit unintuitive. I think we should add this in the future.

Will add those features in the next days (either with this or with a new PR, depending whether gets merged first).

lluchs:
> > Maybe it should also indicate that it's showing a filtered list. Something like "Search results:" on top?
> 
> I find it a bit redundant, as the player can check whether the searchbar's empty. But then again, this might not work if he entered whitespace...

Right, but the important part that is not visible at all is whether the user has already pressed "Enter" for the current search term or not. Did you try search-as-you-type (without pressing "Enter"), maybe? But that will probably need some additional work to have acceptable performance while typing.

src/gui/C4StartupScenSelDlg.cpp
walachey: An additional (classical) example to what Marky said:
`C4Gui::ListBox* left, right;` would declare one pointer and one object on the stack.

In general, we do not (and can't) cover everything with the style guidelines. The guidelines are mainly for the things people usually argue about (e.g. tabs vs. commas, variable names...).

src/gui/C4StartupScenSelDlg.h
walachey: The comments here are not personal criticism :) It doesn't really matter who introduced that code smell, but touching code is a great opportunity to also reduce the technical debt of the project a little bit each time :)

SvenTwo:
Should this be merged?

lluchs:
I merged it.

PR #114, opened 2019-07-04
Refactor clonk

gitMarky:
Split the Clonk script into several parts. Currently, if you create an external pack, you can extend the Clonk as a whole - inheriting a complex bunch of behaviour with it. This may be desirable sometimes, but in many cases it is not. 
The logic is as follows:

- Clonk_Generic contains the basic functionality for a Clonk and/or humanoid, very much like an interface. It handles most of the engine callbacks, except for digging out objects, because that sort of relies in the bucket, which is an OC specific item
- Clonk_HandDisplay contains the logic for displaying items in the Clonk hands with the current left/right hand inventory system. Not necessary if you define an entirely different kind of inventory, for example.
- Clonk_InteractionMenu handles the OC interaction menu. As above, if you have something different you will not need that.
- Clonk_Sounds adds the default sounds *including* the skin sounds. This might be a bit messy, looking forward to your opinions on that.
- Clonk_Skins handles the 4 default skins. If you do not need those, you can skip including them.
- Clonk itself then adds the ActMap (which *might* go to Clonk_Animations?), and some very specific things, such as adding and removing the backpack (in external packs, many times you want to skip the backpack anyway). ActMap can be added to other objects via Definition(), i.e. def.ActMap = Clonk.ActMap or use Clonk.ActMap as the prototype.
- Magic energy was added as a library, instead of a "Clonk_"-flavour, because anything could have magic energy.


772:
Recently, I had problems overloading the Clonk's ActMap via an "#appendto Clonk" script in System.ocg so I had to overload the whole object. That behaviour is not affected, is it?

gitMarky:
>Recently, I had problems overloading the Clonk's ActMap via an "#appendto Clonk" script in System.ocg so I had to overload the whole object. That behaviour is not affected, is it?
I am not sure whether I understand the question.
To my understanding, `#appendto Clonk local ActMap = ...;` will overwrite the ActMap in Clonk entirely. Only if you create a definition with #include Clonk you can modify parts of the ActMap.
If you want to do this, you could 
`#appendto Clonk

func Definition (id type)
{
  _inherited(type, ...);
  type.ActMap = { Prototype = Clonk.ActMap, <your additional actions>}
}` where modified actions could be `Walk = { Prototype = Clonk.Walk, Speed = 1337,}` or whatever.

In the current state this behaviour stays the same, yes. If we were to move the local ActMap; to Clonk_Animations the Clonk would not define an ActMap and you could use your appendto-script, maybe. This is something I'd have to try out.

gitMarky:
Armin, I just checked something: in one of my projects the following worked:

```
#appendto Clonk

local ActMap = {
    AdditionalAction = {...},
};
```
This added the additional action, but left the other actions unchanged.

PR #118, opened 2019-12-27
Fix accident in C4Group

gitMarky:
In commit 99992b6f47 it seems that the changes were added after the break by accident, resulting in dead code. However, this code did not seem to cause any problems, so I am not sure whether we need this chance


walachey:
@SvenTwo You made the original change

SvenTwo:
This was 8 years ago, so I don't remember. Not sure if we have a loop that deletes entries in a search loop; but shouldn't cause trouble either way.

PR #120, opened 2020-03-06
Fix Grammar ("farer" to "further")

psypherium:
I don't know why line 50 changed in StringTblUS.txt, that wasn't intentional


walachey:
Hey, nice catch :)
Shouldn't it be "farther" rather than "further" for physical distances, though? (I am a non-native speaker)

psypherium:
No, both words are used to refer to physical distances. Some pedants may tell you that "farther" is more correct, but I dislike it due to it being a homophone of the word "father". My opinion is that avoiding homophones is preferable wherever possible, even in written text.

It is up to you which one you want to use. "Farther" is probably preferred by more grammarians.

walachey:
Okay, convinced :) Thanks for the contribution!

PR #130, opened 2020-04-12
Fix various warnings

ewanmellor:
This removes declarations that are unused entirely, or adds macro guards to match declarations that are only used in certain configurations.

Fix a format string that did not match its arguments.  This was broken in 102047f (Feb 2013).

Fix erroneous semicolons before method bodies.  These are ignored in Objective C++, they just cause a warning.

Fix deprecation warnings in macOS.  This replaces a number of deprecated enums and bitmasks with their renames.  It replaces the call to Gestalt (deprecated years ago) with a check against NSAppKitVersionNumber.  It also defines GL_SILENCE_DEPRECATION, which silences the warnings about OpenGL being deprecated on macOS (we're not going to switch to Metal yet).


src/graphics/C4DrawGLCtx.cpp
```diff
@@ -25,8 +25,10 @@
 
 #ifndef USE_CONSOLE
 
+#ifdef USE_WGL
```
lluchs, line 28: Looks like this one is used on Linux as well. I'm fixing it before merging...

PR #135, opened 2020-04-13
Fix build of netpuncher and tests on macOS, and add a Travis macOS build

ewanmellor:
StdSchedulerMac needs to be included in libmisc so that the netpuncher
and tests targets can pick it up.  This is conditional on APPLE being set
(unlike the other StdScheduler* files) because the Travis builds for other
OSes do not have the Objective C++ compiler installed.  Also, there is an
earlier cset that implies that this was a problem more generally (515aaf6).

Clang/libc++ on macOS apparently doesn't support quick_exit, even with
-std=c++11.  Use a macro check in AulDeathTest to switch it to regular
exit instead.

Add test build output and autogen directories to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,12 @@ mape_autogen/
 netpuncher_autogen/
 openclonk-server_autogen/
 openclonk_autogen/
+tests/aul_test
+tests/aul_test_autogen/
+tests/gmock_autogen/
+tests/gtest_autogen/
+tests/tests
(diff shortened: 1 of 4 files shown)


lluchs:
Awesome, thanks! I'll try to get the upload to our download page working.

lluchs:
Now the macOS download is available here: https://www.openclonk.org/snapshots/

PR #140, opened 2020-05-05
Fix slowness in deleting proplist references

jcaesar:
```js
// Example script
global func Main() {
    var x = Bla(1000 * 12);
    Log("Done");
}

global func Bla(count) {
    if (count < 1)
        return nil;
    count -= 1;
    var x = new Global { i = count };
    var a = Bla(count / 2);
    var b = Bla(count - count / 2);
    x.b = b;
    x.a = a;
    return x;
}
```

The phenomenon occurs depending on whether the order of creating `a` and `b` is the same as assigning them to `x` or not. On some architectures/compiler/os combinations, the problem may also occur randomly, regardless of the order.

Execution times vary between 15 ms and 700 ms without this patch.

### DIFF
--- a/src/script/C4PropList.cpp
+++ b/src/script/C4PropList.cpp
@@ -24,41 +24,17 @@
 
 void C4PropList::AddRef(C4Value *pRef)
 {
-#ifdef _DEBUG
-	C4Value * pVal = FirstRef;
-	while (pVal)
-	{
-		assert(pVal != pRef);
-		pVal = pVal->NextRef;
-	}
-#endif
-	pRef->NextRef = FirstRef;
-	FirstRef = pRef;
+	assert(Refs.count(pRef) == 0);
+	Refs.insert(pRef);
 }
 
-void C4PropList::DelRef(const C4Value * pRef, C4Value * pNextRef)
+void C4PropList::DelRef(C4Value * pRef)
 {
-	assert(FirstRef);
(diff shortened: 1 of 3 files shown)


lluchs:
LGTM, your example script is clearly faster - although the minimum is a bit higher now. We're clearly missing a benchmark suite to evaluate changes like this one better.

```
hyperfine -- './c4script perf.c' './c4script_old perf.c'
Benchmark #1: ./c4script perf.c
  Time (mean ± σ):      25.4 ms ±   3.3 ms    [User: 22.7 ms, System: 2.5 ms]
  Range (min … max):    21.2 ms …  41.2 ms    118 runs

Benchmark #2: ./c4script_old perf.c
  Time (mean ± σ):     432.0 ms ± 331.8 ms    [User: 421.7 ms, System: 3.9 ms]
  Range (min … max):    10.3 ms … 741.2 ms    211 runs

Summary
  './c4script perf.c' ran
   16.99 ± 13.23 times faster than './c4script_old perf.c'
```

jcaesar:
Hm. On the other hand, memory usage did go up quite a bit, from 12 to 17 MiB (if you increase the number of proplists from 12'000 t to 36'000).

lluchs:
I think that's fine. The average game most likely doesn't have that many proplists anyways?

PR #141, opened 2020-05-05
Extra menu entry in the credits screen for displaying the licenses

jcaesar:
See https://bugs.openclonk.org/view.php?id=2069


jcaesar:
@isilkor Can I ask you to have a look and tell me whether you think this is sufficient?

isilkor:
From a pure reading of the code this seems to fulfill our obligations.

Issue #137, opened 2020-04-20
Game does not start on Wayland


kaimast:
Hi,
I'm using Ubuntu 20.04 on Wayland. Usually X11/GLX games just work fine through XWayland. The binary form the openclonk package however does not start. Also there seems to be no .desktop file, so the only way to run it is via the command line.

```
kai@kai-thinkpad ~> openclonk
Command 'openclonk' is available in '/usr/games/openclonk'
The command could not be located because '/usr/games' is not included in the PATH environment variable.
openclonk: command not found
kai@kai-thinkpad ~ [127]> /usr/games/openclonk
[16:10:31] OpenClonk
[16:10:31] Version: 8.1 unix linux-x86_64 (5c6fadfb266d - 2018-03-16 18:25:33 +0100)
[16:10:31] ExePath: "/usr/games/"
[16:10:31] SystemDataPath: "/usr/share/games/openclonk/"
[16:10:31] UserDataPath: "/home/kai/.clonk/openclonk/"
[16:10:31] Command line: /usr/games/openclonk
[16:10:31] Couldn't find matching GLX visual
[16:10:31] Game cleared.
[16:10:31] Engine shut down.
```

The same problem is present with the flatpak.  
```
kai@kai-thinkpad ~ [1]> flatpak run org.openclonk.OpenClonk
[16:14:34] OpenClonk
[16:14:34] Version: 8.1 unix linux-x86_64 (5c6fadfb266d - 2018-03-16 18:25:33 +0100)
[16:14:34] ExePath: "/app/bin/"
[16:14:34] SystemDataPath: "/app/share/games/openclonk/"
[16:14:34] UserDataPath: "/home/kai/.clonk/openclonk/"
[16:14:34] Command line: openclonk
[16:14:34] Couldn't find matching GLX visual
[16:14:34] Game cleared.
[16:14:34] Engine shut down.
```

lluchs:
Yes, unfortunately, this broke when SDL2 started to use Wayland per default. The game should work with XWayland. You can either unset `WAYLAND_DISPLAY` or set `SDL_VIDEODRIVER=x11` to force running in XWayland. Of course, we should support Wayland eventually, but I personally don't run it on any of my systems yet.

@jcaesar This might be something we can fix in the flatpak by adjusting the environment variables?

>Also there seems to be no .desktop file, so the only way to run it is via the command line.

[...]

PR #142, opened 2020-05-20
SDL window init: test availability of AA/subsampling before requiring it

jcaesar:
No answer in #137 but whatever. It does fix the issue on my machine.

### DIFF
--- a/src/graphics/C4DrawGLMac.mm
+++ b/src/graphics/C4DrawGLMac.mm
@@ -573,7 +573,7 @@ void CStdGLCtx::Clear(bool multisample_change)
 	}
 }
 
-void C4Window::EnumerateMultiSamples(std::vector<int>& samples) const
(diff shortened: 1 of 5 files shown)


lluchs:
Ah, I wanted to try this, but then had unrelated OpenGL issues on my laptop. I'll have another look...

jcaesar:
Ty. I don't have a windows available for testing, either… :/

lluchs:
The patch works on my system. Maybe it should also change invalid configuration values. Currently, it will show 4x MSAA in the options even when it's in fact disabled.

jcaesar:
I made it write the actual multisampling value back to the config.

Good thing you brought it up. Turns out that calling `glGetIntegerv(GL_MAX_SAMPLES, &max_samples);` at that point (probably before creating the first window, but I'm not sure what the reason is) will return 0, even if AA is available. So the original patch would have broken AA for everything.
I added a workaround.

PR #144, opened 2020-07-03
Add French translation by janolap1

lluchs:
Mostly looks fine, but from a quick check it seems to cause some script errors. Will need some testing before merging!

We also need to credit the creators properly, at least with an entry in planet/AUTHORS, as well as the credits line below the language dropdown menu. Maybe also on the Credits page, but I'll update that in bulk before the release.


planet/Missions.ocf/Raid.ocs/System.ocg/StringTblFR.txt
```diff
@@ -0,0 +1,215 @@
...
+Woody3=Pourriez-vous aller vers l'est (à droite) et abattre un conifère? Je m'occuperai alors du sciage.
+Woody4=D'accord. Mais je pense que votre hache est un peu grosse pour un débutant comme moi. Avez-vous une hache normale quelque part ?
+Woody5=Mes outils sont stockés près de Rocky dans la mine Ouest (à gauche) d'ici.
+Woody6=Où puis-je abattre des arbres?
+Woody7=Il suffit de travailler sur l'un des conifères à l'est (à droite) d'ici.
+
+Woody100=Comment pouvons-nous nous opposer à cela ?
+
+
+Rocky1=Salut, avez-vous une hache pour moi ?
```
gitMarky, line 89: The spaces before punctuation signs are inconsistent or I do not understand the pattern. Before the question and exclamation marks there always seems to be a space, but before a full stop there is no space. Except in line 83 the space before the question mark is missing.

PR #145, opened 2020-07-25
Lorry ingame integration

walachey:
This provides an ingame interface to the user-content web frontend "Lorry": https://github.com/walachey/openclonk-lorry/
The web frontend is running at https://mods.openclonk.org/

Ingame, the player can view, search, sort, filter (with filters provided: only playable scenarios, compatible scenarios) available mods and then install them. Installed mods will be put to "%APPDATA%/mods/UUID_modnameslug". The game will search that location when displaying the scenario list.
Installed mods can be checked for updates - the SHA1 checksums of installed files are then compared with the ones provided by the API. If there is a discrepancy, the mod will be updated.

Mods can request dependencies which are then also installed alongside the mod.

I did NOT squash all the commits, because the ingame client is now three years old and the history is even useful for me to check why certain things are the way they are. I could obviously squash some hand picked commits in between if it is worth the effort.

Feedback?


lluchs:
Looks like there are some build errors on Linux with old gcc and macOS.

gitMarky:
Regarding the  https://github.com/walachey/openclonk-lorry/ repo: Is there a manual how to upload mods? Did not find it something. Ok, nevermind, after logging in and refreshing there was an option "Upload new package"

src/config/C4Reloc.cpp
```diff
@@ -69,14 +71,83 @@ bool C4Reloc::AddPath(const char* path, PathType pathType)
...
+		return *temporaryPathInfo;
+	}
+	return *pathListIter;
+}
+
+bool operator==(const C4Reloc::const_iterator& a, const C4Reloc::const_iterator& b)
+{
+	if (!a.subdirIters.empty()) return false;
+	if (!b.subdirIters.empty()) return false;
+	if (a.pathListIter != b.pathListIter)
```
gitMarky, line 130: You could reduce these three lines to `return a.pathListIter == b.pathListIter;`

src/game/C4Application.cpp
```diff
@@ -450,6 +450,19 @@ void C4Application::ParseCommandLine(int argc, char * argv[])
...
+				// IDs are at least a few characters long.
+				const char *id = Game.DirectJoinAddress + 10;
+				if (SLen(id) > 10)
+				{
+					++id; // Remove slash.
+					C4Startup::SetStartScreen("mods", id);
+				}
+				Game.DirectJoinAddress[0] = 0;
+				continue;
+			}
```
gitMarky, line 465: The 10 appears multiple times here - if they all have the same meaning it would be good to use a constant instead of 10. No idea what this magic number is good for, could you add a comment? Looks like some kind of prefix for determining the ID

walachey: Yes, true. I can make that a bit more readable. The 10 ist just strlen("installmod") :)

walachey:
>     * The "show installed" button on the left is a one-way street, no way to go back to "show all".
>     * The "search online" button is a bit confusing. Possibly should be labeled "Refresh"?
The "search online" button is exactly the way to go back to "show all". Hmm. I guess that would be even more confusing if labelled "Refresh"?

>     * I feel like there should be a way to "start" a mod after installing it (whatever that means - maybe opening the "start game" page with the scenario/folder pre-selected?). There's a serious "what now??" moment after installing something.
True. Not sure about that though. What happens if there are e.g. multiple scenarios in a mod?
I could go to the actual folder where the mods are installed so you can only see what is contained there. But that's also a bit misleading (because usually you will select the mod from the normal scenario list).

>     * Would it be possible to add some UI for mods in the "start scenario" UI? Something like "This is from mod XYZ, click here to go to the mods page".
"Go to the mods page" would mean open an external website?
Would probably be possible. Only that currently the scenario list does not attempt to parse the XML data of the mods. So that might make loading noticably slower if many mods are installed. hmm

>     * Hovering over the star tag shows broken translation
>     * Typo "lokal" on the Remove button tooltip
Fixed

>     * The filters on the right should probably also show the corresponding icons from the list.
Good idea

>     * Will everything break if someone uploads something with a SHA-1 collision? :P
The files are held in separate folders for each mod. And when checking files, it will first check if it finds the file with that name in the mod's folder and then check its checksum. So two files from two different mods don't interfere with each other in any way (and can also even have the exact same filename). Not sure what happens if you upload two files with identical names in ONE mod (e.g. upload them on case-sensitive linux and then download them on windows) though

> Looks like there are some build errors on Linux with old gcc and macOS.

Meh, I'll check. Maybe it's enough to replace the `auto` keyword there.

lluchs:
>The "search online" button is exactly the way to go back to "show all". Hmm. I guess that would be even more confusing if labelled "Refresh"?

Ah. I'd expect the "show installed" button to behave like a filter toggle, so that clicking it twice would go back to the previous view. Apparently that's not what it actually does?

>I could go to the actual folder where the mods are installed so you can only see what is contained there. But that's also a bit misleading (because usually you will select the mod from the normal scenario list).

The other issue is whether it would start a normal game or open a lobby.

Maybe there could be an alternate scenario list mode that shows only mod folders. That could reduce confusion from a completely different view on the mods page?

>"Go to the mods page" would mean open an external website?

No, I meant going to the in-game mods UI from the scenario list so that you can check for updates etc. A "go to website" button might make sense for the mods page, though, to simplify discovery ("how do I upload my own mod?").

>Would probably be possible. Only that currently the scenario list does not attempt to parse the XML data of the mods. So that might make loading noticably slower if many mods are installed. hmm

Parsing a single XML file for a mod should be pretty quick, I suppose? Maybe it could be done on demand when clicking on a scenario. I think the whole "pre-load everything" approach of the scenario list is only because it needs to extract title and icon anyways to display the list itself.

walachey:
> Ah. I'd expect the "show installed" button to behave like a filter toggle, so that clicking it twice would go back to the previous view. Apparently that's not what it actually does?

No, clicking the "Show installed" on the left goes through your local mods folder and lists everything it finds there. That's also (by design) completely independent of the information provided by the API. So even if a mod is deleted online, it will show there. Even if the server is offline, it will show there. Even if you have manually added a folder inside your "mods" folder that has nothing to do with anything installed by Lorry, it will show there.
"Search online" is the opposite, sending an API request with all the filter options and displaying the results (and, in the results, marking things as "Installed" if the UUID is found in your mods folder).

> Maybe there could be an alternate scenario list mode that shows only mod folders. That could reduce confusion from a completely different view on the mods page?

I am still wondering if all that would have the opposite effect and make it more complicated to understand what happens. (E.g. if there is a button that goes to a "mod only" view, then a user might ask themselves if the mods are also shown in the normal view or whether he has to go through the additional click of going to the "mod only" list)

gitMarky:
Could you add your explanation as tooltips to the buttons? It is really helpful!

walachey:
> Could you add your explanation as tooltips to the buttons? It is really helpful!

What would that mean? The tooltips are atm `Displays all locally installed mods.` and `Search the online database according to the search query and filters (hotkey F5).`

gitMarky:
> > Could you add your explanation as tooltips to the buttons? It is really helpful!
> 
> What would that mean? The tooltips are atm `Displays all locally installed mods.` and `Search the online database according to the search query and filters (hotkey F5).`
The tooltip for "Search online" is good as it is. For "Show installed" I'd expand it, since your explanation to Luchs has cleared things up: "Displays all locally installed mods. The list is independent of the online database." or even more "... of the online database and works if the database is offline. It may also contain mods that do not exist online, either because of deletion from the database or because they are new." Maybe I am overthinking it :D

walachey:
I think you are overthinking it :)
Those are very interesting details for developers, but a tooltip should help a player understand /what/ the things do and not how the technical infrastructure behind that works.
Might be better suited for a blog post about it

lluchs:
Linux is building now, the macOS build still fails with:

```
/Users/travis/build/openclonk/openclonk/src/platform/C4AppDelegate.mm:236:66: error: cannot increment value of type 'C4Reloc::iterator' (aka 'C4Reloc::const_iterator')
        for (C4Reloc::iterator it = Reloc.begin(); it != Reloc.end(); it++)
                                                                      ~~^
/Users/travis/build/openclonk/openclonk/src/platform/C4AppDelegate.mm:243:14: error: member reference type 'C4Reloc::iterator' (aka 'C4Reloc::const_iterator') is not a pointer; did you mean to use '.'?
                else if (it->pathType == C4Reloc::PATH_PreferredInstallationLocation)
                         ~~^~
                           .
/Users/travis/build/openclonk/openclonk/src/platform/C4AppDelegate.mm:243:16: error: no member named 'pathType' in 'C4Reloc::const_iterator'
                else if (it->pathType == C4Reloc::PATH_PreferredInstallationLocation)
                         ~~  ^
/Users/travis/build/openclonk/openclonk/src/platform/C4AppDelegate.mm:244:45: error: member reference type 'C4Reloc::iterator' (aka 'C4Reloc::const_iterator') is not a pointer; did you mean to use '.'?
                        destPath = [NSString stringWithCString:it->strBuf.getData() encoding:NSUTF8StringEncoding];
                                                               ~~^~
                                                                 .
/Users/travis/build/openclonk/openclonk/src/platform/C4AppDelegate.mm:244:47: error: no member named 'strBuf' in 'C4Reloc::const_iterator'
                        destPath = [NSString stringWithCString:it->strBuf.getData() encoding:NSUTF8StringEncoding];
                                                               ~~  ^
```

Two new ideas regarding the UI:

Would it make sense to move the "Search online" button to the left column so that "Search online" and "Show installed" behave a bit like radio buttons? Does the rest of the UI (like searching and filtering) work while "Show installed" is active?

A simpler alternative to solve the "what now" problem after installing a mod might be to show a message box with the mod contents when clicking an installed mod. Something like (not sure whether to include objects):

```
Mod Gold Rush Extreme contains:
 * [scenario icon] Gold Rush Extreme
 * [object icon] CaveSpiders
You can find these [scenario icon][scenario folder icon] in the scenario list!
```

Would require reading the ocs/ocf files to retrieve the localized title, though. Otherwise, the list wouldn't be of much help.

walachey:
> Does the rest of the UI (like searching and filtering) work while "Show installed" is active?

Nope, would also be a hassle to implement a similarly flexible search from scratch (as I am e.g. using the grammar compression from postgres (allowing to find the word "Wipf" when looking for "wipfs").
Also, it's not "active". It's just a one-shot clickable button. Maybe that's confusing because it doesn't look like the other buttons? I could replace the image by a normal button.
Then again, is it really that confusing or is it something that's understood easily once you click the button once or twice?

Regarding the "what now" thing. As it's probably only confusing the first time, might be enough to add a little help text below the "Show installed" button? (something like "you will find all scenarios from installed mods in the Start Game list")

gitMarky:
Changes look good, merge from master looks good, too!

lluchs:
>Also, it's not "active". It's just a one-shot clickable button. Maybe that's confusing because it doesn't look like the other buttons? I could replace the image by a normal button.

Okay, but could be made active (I think the C4GUI buttons have some brighter background for that)? The view is definitely active after it's clicked, until you interact with anything else, I guess.

>once you click the button once or twice?

My whole point is that clicking it twice doesn't do anything, while I'd expect something to happen ;)

>might be enough to add a little help text below the "Show installed" button?

Maybe - is it always clear what to look for in the scenario list? I believe the "mod name" and scenario/folder title are currently completely disconnected, and there's no localization for the mod name. So we'd need the "mod name" to also show up in the scenario list.

lluchs:
> Okay, I think it's fine now for merging to get more feedback from others. I don't really care about the code style stuff - there's a lot more places where `*` and `&` are inconsistently placed.

walachey:
Okay, squashed a lot of the commits where it made sense.

PR #152, opened 2021-01-03
openclonk.desktop: include %u replacement to denote support for URLs

pkern:
It is recommended to include a notion of how to pass arguments to the application when MIME handlers are listed as supported. %u denotes a single URL, which seems to match openclonk's documentation. Multiple URLs would then be opened in separate applications.


SvenTwo:
Sounds good. I assume this is so you can e.g. draG+drop a scenario file onto the shortcut to launch it (like under Windows)?

pkern:
That was unclear to me. If so, it should likely also cope with file:// URLs in addition to filepaths. It looks like the primary purpose of (open)clonk:// URLs was to support sites like https://clonkspot.org/games and let those links auto-open these linked games.

PR #160, opened 2023-01-30
Add missing #include directives needed for recent GCC versions

musicinmybrain:


lluchs:
Thanks. I think this has been fixed otherwise, since building current master seems to work fine.

Issue #163, opened 2023-06-11
Fail to build on Arch


jpretori:
Installing 8.1 from the AUR, fails with a build error.

It turns out this is related to newer versions of GCC defaulting to NOT automagically linking in limits.h.

This is fixable by a very simple edit to /etc/makepkg.conf (just add  "-include /usr/include/c++/13*/limits" to CXXFLAGS).

I'll submit a PR for a small update to the README file to make people aware.

PR #164, opened 2023-06-11
Update README with notes for Arch builds

jpretori:
Fixes issue #163 by making people aware that limits.h not being included automatically is the problem, and giving them a simple and effective fix for that, that will persist for future releases of openclonk.


Fulgen301:
Since this is an OpenClonk PR as opposed to an AUR notice, this is a hack for a bug that should be properly fixed by including`<climits>` (or better yet, using `<limits>`) in the relevant files.

lluchs:
Thanks. I verified that building current master works fine on Arch Linux.

PR #166, opened 2024-05-16
CMakeLists: fix forced value for WITH_APPDIR_INSTALLATION

lheckemann:


lluchs:
I think the idea here was that if you turn on `WITH_AUTOMATIC_UPDATE` (which defaults to OFF), you want website builds, which are AppImages. Other installation modes (distribution packages, flatpak) update externally and wouldn't turn on `WITH_AUTOMATIC_UPDATE`.

lheckemann:
Right, but the previous code [forced](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) `WITH_APPDIR_INSTALLATION` to `ON` if the conditions weren't met (i.e. automatic updates were turned off), which is exactly the wrong way round.

lheckemann:
This meant that for example the Nix package would be built with `WITH_APPDIR_INSTALLATION` enabled (and impossible to disable), which meant that SystemDataDir was set incorrectly and the game data couldn't be found.

lluchs:
Yeah, that was stupid. Thanks!

PR #171, opened 2026-04-28
Fix appdata paper cuts

ghost:
Partly, based on: 

https://github.com/flathub/org.openclonk.OpenClonk/blob/master/appstream.diff


lluchs:
Thanks!

